PR #4, opened 2016-09-09
Move document observer to onReady.

kevinpschaaf:
Needed to ensure deterministic ordering of upgrades when loading HTMLImports polyfill.


sorvell:
LGTM

PR #9, opened 2016-09-09
Make sure to avoid calling native `createElement` with a second argum…

sorvell:
…ent if none is needed. Previously `undefined` was passed and this causes Firefox to create and element with `is="undefined"`, which is obviously undesirable.


kevinpschaaf:
LGTM

ebidel:
Will we also be adding this to the v0 polyfills? @dglazkov promised that in https://github.com/w3c/webcomponents/issues/544#issuecomment-241542907 :)

PR #11, opened 2016-09-12
Better compatibility with the HTMLImports polyfill

sorvell:
Always batch upgrades to occur at `HTMLImports.whenReady` time.

This is necessary for two reasons:

1. correctness: Customization *should* occur as elements are parsed and are thus ordered with script execution. While this has been punted for main document, this change helps enforce this rule for HTMLImports by synchronizing upgrades with HTMLImports ready state. Without it and because elements in imports are upgraded after the import is loaded, if in an async loaded import a script defines a custom element A that depends on a custom element B in the import, upgrades for element of type A will fail because B has not yet upgraded. 
1. performance: currently a full document walk is performed for each call to define. By synchronizing this with the ready state of HTMLImports, elements defined in imports are batched and cause only 1 walk. An alternative approach to batching would be to walk once and keep a map of upgrade candidates (I believe the platform uses this approach).


justinfagnani:
If there's a correctness issue here, it's pretty critical to have some sort of test that was failing before the change, and I have no way of knowing if I revert the behavior.

justinfagnani:
LGTM

PR #16, opened 2016-10-06
Update native-shim.js to not use new.target

justinfagnani:
Do not merge. For discussion.

PR #22, opened 2016-10-24
Update native-shim.js

tuespetre:
Wrap the shim in a function constructor within a try/catch block so it can be safely included when desired.


tuespetre:
I just now saw #16. Oh well, I will leave this here for someone else to close if they want 😄

snuggs:
#ping @justinfagnani can we close this?

justinfagnani:
The new shim has the same issue because `class` is a syntax error in older browsers, but this PR is outdated anyway.

PR #23, opened 2016-11-02
New shim w/o Reflect.construct

justinfagnani:


src/native-shim.js
```diff
@@ -18,15 +18,125 @@
...
+   * @type {Map<string, Function>}
+   */
+  const constructorByTagname = new Map();
+
+
+  // /**
+  //  * Whether the constructors are being called by a browser process, ie parsing
+  //  * or createElement.
+  //  */
+  // let browserConstruction = false;
```
justinfagnani: Commented out so I can expose to test some pre and post conditions. I'll move back to these private variables soon.

src/native-shim.js
```diff
@@ -18,15 +18,125 @@
...
+          // Make sure that user-defined constructor bottom's out to a do-nothing
+          // HTMLElement() call
+          HTMLElement.browserConstruction = true;
+          // Call the user-defined constructor on our instance:
+          elementClass.call(this);
+        }
+        HTMLElement.userConstruction = false;
+      }
+    }
+    if ('observedAttributes' in elementClass) {
```
justinfagnani: I need tests for the reactions

src/native-shim.js
```diff
@@ -18,15 +18,125 @@
...
+   */
+  HTMLElement.browserConstruction = false;
+
+  /**
+   * Whether the constructors are being called by a user-space process, ie
+   * calling an element constructor.
+   */
+  HTMLElement.userConstruction = false;
+
+  window.customElements.define = (tagname, elementClass) => {
```
justinfagnani: It'd be nice to check that elementClass is an ES5-style class and throw if not. This approach requires that all constructors in a chain are ES5 - that the bottom-most constructor can be called with `Function.call()`

src/native-shim.js
```diff
@@ -10,23 +10,161 @@
...
+ *  Compiling valid class-based custom elements to ES5 will satisfy these
+ *  requirements with the latest version of popular transpilers.
  */
 (() => {
-  let origHTMLElement = HTMLElement;
-  // TODO(justinfagnani): Tests!!
+  'use strict';
+
+  const NativeHTMLElement = HTMLElement;
+  const nativeDefine = window.customElements.define;
```
ajklein: Why do you access off window for customElements but not for HTMLElement?

src/native-shim.js
```diff
@@ -10,23 +10,161 @@
...
+      return instance;
+    }
+    // Else do nothing. This will be reached by ES5-style classes doing
+    // HTMLElement.call() during initialization
+    browserConstruction = false;
+  };
+
+  HTMLElement.prototype = Object.create(NativeHTMLElement.prototype);
+  HTMLElement.prototype.constructor = HTMLElement;
+  HTMLElement.nativeClass = NativeHTMLElement;
```
ajklein: What's this about? I only see it used in tests.

src/native-shim.js
```diff
@@ -10,23 +10,161 @@
...
+
+  window.customElements.define = (tagname, elementClass) => {
+
+    const StandInElement = class extends NativeHTMLElement {
+      constructor() {
+        // Call the native HTMLElement constructor, this gives us the
+        // under-construction instance as `this`:
+        super();
```
ajklein: I take it there are never arguments to pass to HTMLElement (I don't know v1 too well)?

src/native-shim.js
```diff
@@ -10,23 +10,161 @@
...
+   * @type {Map<string, Function>}
+   */
+  const constructorByTagname = new Map();
+
+
+  /**
+   * Whether the constructors are being called by a browser process, ie parsing
+   * or createElement.
+   */
+  let browserConstruction = false;
```
ajklein: Is it not possible for these to get confused if nesting happens? I haven't thought it through, but it seems like these are instance-specific...

src/native-shim.js
```diff
@@ -10,23 +10,161 @@
...
+        if (!userConstruction) {
+          // Make sure that user-defined constructor bottom's out to a do-nothing
+          // HTMLElement() call
+          browserConstruction = true;
+          // Call the user-defined constructor on our instance:
+          elementClass.call(this);
+        }
+        userConstruction = false;
+      }
+    }
```
ajklein: Nit: semicolon here (and below wherever you use a function expression)

src/native-shim.js
```diff
@@ -10,23 +10,161 @@
...
+    if ('observedAttributes' in elementClass) {
+      StandInElement.observedAttributes = elementClass.observedAttributes;
+    }
+    const elementProto = elementClass.prototype;
+    if ('connectedCallback' in elementProto) {
+      StandInElement.prototype.connectedCallback = function() {
```
ajklein: Is it important that this is a new function, rather than just copying connectedCallback off the old proto? Same goes for all the callbacks below.

src/native-shim.js
justinfagnani: Oops, it is only for tests. I'll remove

src/native-shim.js
justinfagnani: Nesting is specifically disallowed in the CE spec, but I should test/guard.

src/native-shim.js
justinfagnani: correct

src/native-shim.js
justinfagnani: Done

src/native-shim.js
justinfagnani: fixed, going with `window.`

src/native-shim.js
justinfagnani: Uh, you're right. I initially thought I needed to set `this` to something, but I clearly don't. @sorvell says this change gets us to within 10% of non-shimmed perf!

src/native-shim.js
```diff
@@ -10,23 +10,161 @@
...
+      }
+    }
+    if ('disconnectedCallback' in elementProto) {
+      StandInElement.prototype.disconnectedCallback = function() {
+        elementProto.disconnectedCallback.call(this);
+      }
+    }
+    if ('attributeChangedCallback' in elementProto) {
+      StandInElement.prototype.attributeChangedCallback = function() {
+        elementProto.attributeChangedCallback.call(this);
```
justinfagnani: Added test. Arguments fixed by just copying functions.

src/native-shim.js
```diff
@@ -10,23 +10,140 @@
...
+      const instance = new (fakeClass)();
+      return instance;
+    }
+    // Else do nothing. This will be reached by ES5-style classes doing
+    // HTMLElement.call() during initialization
+    browserConstruction = false;
+  };
+
+  window.HTMLElement.prototype = Object.create(NativeHTMLElement.prototype);
+  window.HTMLElement.prototype.constructor = window.HTMLElement;
```
ajklein: Nit: you can use NativeHTMLElement on the RHS here.

PR #24, opened 2016-11-02
Update readme to mention native-shim

robdodson:
@sorvell PTAL

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,3 +52,7 @@ customElements.flush();
 
 // some more DOM operations dependent on reactions in the first set
 ```
+
+## Known Issues
+
+Currently compiling Custom Elements down to ES5 will result in issues in browsers that already natively support Custom Elements as classes. We're working on a fix which should be forthcoming soon. In the meantime you can include the [native-shim.js](https://github.com/webcomponents/custom-elements/blob/master/src/native-shim.js).
--- a/README.md
+++ b/README.md
@@ -55,4 +55,8 @@ customElements.flush();
 
 ## Known Issues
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,3 +52,7 @@ customElements.flush();
 
 // some more DOM operations dependent on reactions in the first set
 ```
+
+## Known Issues
+
+Currently compiling Custom Elements down to ES5 will result in issues in browsers that already natively support Custom Elements as classes. We're working on a fix which should be forthcoming soon. In the meantime you can include the [native-shim.js](https://github.com/webcomponents/custom-elements/blob/master/src/native-shim.js).
```
justinfagnani: The fix we are working on is simply a faster native-shim.js

sorvell: I would say something like...

"The Custom Elements v1 spec is not compatible with ES5 style classes. This means ES6 code compiled to ES5 will not work with a native implementation of Custom Elements.* While it's possible to force the custom elements polyfill to be used to workaround this issue (by setting (`customElements.forcePolyfill = true`), you will obviously not be using the browser's native implementation in that case.

Since this is not ideal, we've provided an alternative: `native-shim.js`. Loading this shim minimally augments the native implementation to be compatible with ES5 code. We are also working on some future refinements to this approach that will improve the implementation and automatically detect if it's needed.

`*` The spec requires that an element call the HTMLElement constructor. You might do this in an ES5 style class like this `HTMLElement.call(this)`. However, that code results in an exception (on Chrome): "Uncaught TypeError: Failed to construct 'HTMLElement': Please use the 'new' operator, this DOM object constructor cannot be called as a function."

robdodson:
updated to include @sorvell's comments

justinfagnani:
LTGM

PR #33, opened 2016-12-09
Fix instanceof for built-in elements.

justinfagnani:
Thanks @kevinpschaaf!


tests/js/instanceof.js
```diff
@@ -20,43 +20,43 @@ suite('Built-in Element instanceof', function() {
...
       'audio',
-      // 'b',
+      'b',
       'base',
-      // 'bdi',
-      // 'bdo',
-      // 'blockquote', // fails on IE11
+      'bdi',
+      'bdo',
+      'blockquote', // fails on IE11
```
kevinpschaaf: presumably these "fails" comments can be removed, assuming they were due to differences in constructors between browsers?

justinfagnani: done

PR #35, opened 2016-12-14
Create `debug` and `rollup` tasks for debugging.

bicknellr:
The 'debug' task is the same as 'default', with extra options given to closure compiler to make its output a bit easier to read.
The 'rollup' task is used to compile modules only and leave as much unmodified as possible. I think this will be easiest to debug, but it's furthest from the compiled form.


gulpfile.js
```diff
@@ -13,22 +13,48 @@
...
+      language_out: 'ECMASCRIPT5_STRICT',
+      externs: ['externs/html5.js', 'externs/custom-elements.js'],
+      dependency_mode: 'STRICT',
+      entry_point: ['/src/custom-elements'],
+      js_output_file: 'custom-elements.min.js',
+      new_type_inf: true,
+      rewrite_polyfills: false,
+    };
+
+    return gulp.src('./src/*.js', {base: './'})
```
bicknellr: Maybe this should be `'./src/**/*.js'`?

gulpfile.js
```diff
@@ -13,22 +13,48 @@
...
-        }))
+      .pipe(closureCompiler(Object.assign(defaultClosureOptions, closureOptions)))
       .pipe(sourcemaps.write('/'))
       .pipe(gulp.dest('./'));
+  }
+}
+
+gulp.task('default', createDefaultTask());
+gulp.task('debug', createDefaultTask({
+  compilation_level: 'SIMPLE',
```
justinfagnani: why compile at all for debugging?

gulpfile.js
```diff
@@ -13,22 +13,48 @@
...
+  formatting: 'PRETTY_PRINT',
+  debug: true,
+}));
+
+gulp.task('rollup', () => {
+  return rollup({
```
justinfagnani: can this be added when it's needed?

gulpfile.js
bicknellr: The idea was to have `debug` be the same as `default` without name minification (but with renaming) and `rollup` be just enough to de-module-ify everything and change as little else as possible. Maybe `rollup` should become `debug` and this one should be removed?

gulpfile.js
bicknellr: (from comment above) I think this might be the preferred 'debug' task if the one above doesn't really make sense.

bicknellr:
Closing; going to put this change in the same PR that adds modules.

PR #36, opened 2016-12-14
Updates to get the tests passing again.

bicknellr:
- Stops the native shim and its tests from running if `window.customElements` does not exist.
- Tests forcing the polyfill now no longer expect `customElements` to exist.
- Tests now import a Promise polyfill.
- Updated WCT to 6.0.0-prerelease.3 to get ES2015->ES5 compilation during tests. (6.0.0-prerelease.4 seems to have a bug that prevents (at least) local testing.)
  - Removed use of `String#endsWith`, which doesn't get compiled during testing...
- `npm install` now causes `bower install`.
- A test checking that the `HTMLElement#constructor` descriptor has both `writable` and `configurable` true was changed to check that these values match the native descriptor's values.


bicknellr:
I'm still trying to figure out why `wct --sauce ..` isn't working for me, so I can't yet confirm that this is working everywhere.

src/native-shim.js
```diff
@@ -61,6 +61,9 @@
 (() => {
   'use strict';
 
+  // Do nothing if `customElements` does not exist.
+  if (!window.customElements) return;
```
justinfagnani: I think this should check for polyfilled also

tests/html/imports.html
```diff
@@ -12,6 +12,7 @@
   <head>
     <title>Custom Elements: imports integration</title>
     <script>
+      window.customElements = window.customElements || {};
```
justinfagnani:
```
window.customElements = window.customElements || {forcePolyfill: true};
```
?

Only cause this snippet might be what we recommend to users...

src/native-shim.js
bicknellr: Isn't that also a case where you wouldn't want the native shim to run?

tests/html/imports.html
bicknellr: If we do that, then `forcePolyfill` doesn't get added in the case that `customElements` already exists (native).

bicknellr: How about
`(window.customElements = window.customElements || {}).forcePolyfill = true;` ?

bicknellr: (I went ahead and changed to this.)

tests/js/patching.js
```diff
@@ -13,9 +13,11 @@ suite('patching', function() {
   suite('HTMLElement', function () {
 
     test('constructor is configurable and writable', function() {
```
bicknellr: This test doesn't really make sense. On [this line](https://github.com/webcomponents/custom-elements/blob/36451e1dbe4e3b77d98499fc12132573d91d4a8b/src/custom-elements.js#L679), the new `HTMLElement.prototype` is set to the old one, so the descriptor is exactly the same. So, I'm going to remove this.

bicknellr:
@justinfagnani Aside from a couple of unsupported browsers erroring out completely in WCT, I think this is ready for review.

package.json
```diff
@@ -16,14 +16,16 @@
     "url": "https://github.com/webcomponents/custom-elements/issues"
   },
   "scripts": {
+    "install": "$(npm bin)/bower install",
```
joeldenning: This causes things to break for me if I `npm install @webcomponents/custom-elements`. Since bower is a devDependency, it is not installed but the `install` script still tries to use it.

What is the purpose of this `install` script? When somebody is `npm install`ing this project, they don't need the bower_components, do they?

PR #41, opened 2017-01-09
Add gulp task to minify native-shim.js

AlexanderOtavka:
I used babili, since the closure compiler doesn't support outputting es6.

I put the build outputs in separate commits in case you guys don't want to merge those yet, for whatever reason.


AlexanderOtavka:
It looks like #36 needs to get merged before this can.

TimvdLippe:
I am not sure if this is something we want to pursue. The polyfill is intended for older browsers, in particularly IE as the evergreen browsers will get support (someday). @webcomponents/polyfills-eng any opinion on this PR?

AlexanderOtavka:
At this point, this PR is so old it probably isn't even worth fixing the merge conflicts.

justinfagnani:
Also, we minimize in the webcomponents repo

PR #42, opened 2017-01-09
Implementing customized built-in elements

### LINKED ISSUE #13: Extending elements support
After trying to follow [this article](https://developers.google.com/web/fundamentals/primers/customelements/) I found that extend is [not currently supported](https://github.com/webcomponents/custom-elements/blob/master/src/custom-elements.js#L217). Am I missing something? Are there any estimates on when it will be? 
This seems very important but it's not even mentioned in https://github.com/webcomponents/custom-elements#differences-from-spec.

joeldenning:
This fixes #13. This implementation is hopefully very close to the spec for customized built-in elements.
*Note that I based my branch off of https://github.com/webcomponents/custom-elements/pull/36 because the tests are failing on the master branch and I didn't want to duplicate the work that @bicknellr has already done in fixing the tests. This means that #36 should probably be merged in before this pull request.*


bower.json
```diff
@@ -16,7 +16,8 @@
   "license": "BSD",
   "ignore": [],
   "devDependencies": {
-    "web-component-tester": "^4.0.1",
-    "webcomponentsjs": "webcomponents/webcomponentsjs#^0.7.22"
+    "web-component-tester": "6.0.0-prerelease.3",
+    "webcomponentsjs": "webcomponents/webcomponentsjs#v1",
+    "es6-promise": "stefanpenner/es6-promise#^4.0.0"
```
joeldenning: These changes are from #36

package.json
```diff
@@ -16,14 +16,16 @@
...
     "test": "wct"
   },
   "homepage": "http://webcomponents.org",
   "devDependencies": {
+    "bower": "^1.8.0",
     "google-closure-compiler": "^20160822.1.0",
     "gulp": "^3.8.8",
     "gulp-sourcemaps": "^1.6.0",
-    "web-component-tester": "^4.0.1"
+    "web-component-tester": "6.0.0-prerelease.3"
```
joeldenning: All changes to package.json are from #36

src/custom-elements.js
```diff
@@ -138,7 +138,7 @@ let Deferred;
       /** @private {!Map<string, !CustomElementDefinition>} **/
       this._definitions = new Map();
 
-      /** @private {!Map<Function, string>} **/
+      /** @private {!Map<Function, !CustomElementDefinition>} **/
```
joeldenning: I have changed `this._constructors` to so that it has the entire CustomElementDefinition, since we need more than the tagName now when looking up custom elements by their constructor

src/custom-elements.js
```diff
@@ -182,17 +182,11 @@ let Deferred;
...
-      //
-      // It doesn't appear possible to check this condition from script
-
-      // 3:
+      // 2:
       const nameError = checkValidCustomElementName(name);
       if (nameError) throw nameError;
 
-      // 4, 5:
+      // 3:
```
joeldenning: The numbering has changed in the spec for `customElements.define`, so I updated it.

src/custom-elements.js
```diff
@@ -201,25 +195,44 @@ let Deferred;
...
+      // 5:
       /** @type {string} */
-      const localName = name;
+      let localName = name;
+
+      // 6:
+      let _extends = options ? options.extends : null;
+
+      // 7:
+      if (_extends) {
```
joeldenning: This is stuff specific to customizedBuiltInElements

src/custom-elements.js
```diff
@@ -265,14 +273,14 @@ let Deferred;
         observedAttributes: observedAttributes,
       };
 
-      // 16:
-      this._definitions.set(localName, definition);
-      this._constructors.set(constructor, localName);
+      // 12:
+      this._definitions.set(name, definition);
+      this._constructors.set(constructor, definition);
```
joeldenning: See comment above about how the _constructors map has changed

src/custom-elements.js
```diff
@@ -465,7 +476,7 @@ let Deferred;
       visitedNodes.add(element);
 
       /** @type {?CustomElementDefinition} */
-      const definition = this._definitions.get(element.localName);
+      const definition = getDefinition(this._definitions, element);
```
joeldenning: Getting the definition for an element is now more complicated than it used to be -- we need to check if the tagName is a custom element but we also need to check if the `is=""` is a custom element. Since that logic is reused in several places, I extracted this out into the `getDefinition` function

src/custom-elements.js
```diff
@@ -635,6 +646,21 @@ let Deferred;
     }
   }
 
+  /**
+   * @param {!Map<string, !CustomElementDefinition>} definitions
+   * @param {!Node|!HTMLElement|null} node
```
joeldenning: I do not know much about Closure compiler's type system, but here's my attempt. This line of code causes warnings because Closures definition for `Node` doesn't have a `tagName` nor a `is` property.

Do you need me to add types to everything? Also do you need me to fix all type warnings? If so I would appreciate help on doing so

src/custom-elements.js
```diff
@@ -635,6 +646,21 @@ let Deferred;
...
+   */
+  function getDefinition(definitions, node) {
+    const name = typeof node.is === 'string' ? node.is : node.tagName.toLowerCase();
+    const definition = definitions.get(name);
+    if (definition) {
+      // Make sure local name matches the actual node's tagName
+      return definition.localName === node.tagName.toLowerCase() || definition.localName === node.is ? definition : null;
```
joeldenning: We want to return null in the following case:

```js
customElements.define('x-foo', class XFoo extends HTMLButtonElement {}, {extends: 'button'});
document.createElement('div', {is: 'x-foo'}); // This is not a button so the element is not upgraded to an x-foo.
```

See https://github.com/webcomponents/custom-elements/pull/42/files#diff-ba6f9bc9f409ee1e78139fbd7d15902eR283

src/custom-elements.js
```diff
@@ -646,37 +672,42 @@ let Deferred;
   CustomElementRegistry.prototype['_observeRoot'] = CustomElementRegistry.prototype._observeRoot;
   CustomElementRegistry.prototype['_addImport'] = CustomElementRegistry.prototype._addImport;
 
-  // patch window.HTMLElement
+  patchElement('HTMLElement')
```
joeldenning: We patch HTMLElement and all of its subclasses so that users can subclass them

src/custom-elements.js
```diff
@@ -646,37 +672,42 @@ let Deferred;
...
 
-  /** @const */
-  const origHTMLElement = window.HTMLElement;
-  CustomElementRegistry.prototype['nativeHTMLElement'] = origHTMLElement;
-  /**
-   * @type {function(new: HTMLElement)}
-   */
-  const newHTMLElement = function HTMLElement() {
-    const customElements = _customElements();
+  function patchElement(varName) {
```
joeldenning: This function is a generic function that is called with HTMLElement, HTMLButtonElement, etc. The code is almost exactly the same as it was before, except generic. All differences have been called out in review comments below.

src/custom-elements.js
```diff
@@ -646,37 +672,42 @@ let Deferred;
...
-    throw new Error('Unknown constructor. Did you call customElements.define()?');
+        // If there's an being upgraded, return that
+        if (customElements._newInstance) {
+          const i = customElements._newInstance;
+          customElements._newInstance = null;
+          return i;
+        }
+        if (this.constructor) {
+          // Find the tagname of the constructor and create a new element with it
+          const constructorInfo = customElements._constructors.get(this.constructor);
```
joeldenning: We need more info than just the `tagName`, so that we know if we're creating a customized built-in element or not. See comment above about how `_constructors` now has a full definition instead of a tagName

src/custom-elements.js
```diff
@@ -646,37 +672,42 @@ let Deferred;
...
+          const options = constructorInfo.name ? {is: constructorInfo.name} : undefined;
```
joeldenning: This decides whether to call `document.createElement('div', {is: 'x-foo'})` or just `document.createElement('div')`

src/custom-elements.js
```diff
@@ -697,9 +728,19 @@ let Deferred;
    */
   function _createElement(doc, tagName, options, callConstructor) {
     const customElements = _customElements();
+    let isAttr;
+    if (options && options.is) {
+        // We're going to take care of setting the is attribute ourselves
+        isAttr = options.is;
+        delete options.is;
```
joeldenning: Delete `options.is` So that the native browser implementation doesn't freak out

src/custom-elements.js
```diff
@@ -697,9 +728,19 @@ let Deferred;
...
+    }
     const element = options ? _nativeCreateElement.call(doc, tagName, options) :
       _nativeCreateElement.call(doc, tagName);
-    const definition = customElements._definitions.get(tagName.toLowerCase());
+    if (isAttr) {
+        element.setAttribute('is', isAttr);
+        element.is = isAttr;
```
joeldenning: Set the `is` attribute and property on the newly created node

src/native-shim.js
```diff
@@ -61,6 +61,9 @@
 (() => {
   'use strict';
 
+  // Do nothing if `customElements` does not exist.
+  if (!window.customElements) return;
```
joeldenning: This is from #36

tests/js/reactions.js
```diff
@@ -181,6 +246,72 @@ suite('Custom Element Reactions', function() {
 
   });
 
+  suite('customized built-in elements', function() {
+    test('inherits native behavior such as a button\'s disabled property', function(done) {
```
joeldenning: Tests that you can actually get the native behavior for specific elements (like `disabled` for buttons) to work just like it's native.

tests/js/reactions.js
```diff
@@ -181,6 +246,72 @@ suite('Custom Element Reactions', function() {
...
+      button.click();
+
+      setTimeout(function() {
+        isDisabled = false;
+        button.disabled = false;
+        button.click();
+      });
+    });
+
+    test('does not upgrade elements that don\'t match what extends was at definition', function() {
```
joeldenning: This is something I am actually not 100% sure on in the spec. It seems like you should not be able to `<div is="x-foo" />` if you say `customElements.define('x-foo', XFoo, {extends: 'span'})`. But I was not able to find a definitive answer in the spec. Am happy to change if I'm wrong on this.

domenic: This appears to be correct. Spec breadcrumbs: https://html.spec.whatwg.org/#concept-try-upgrade will give a null definition in step 1 since https://html.spec.whatwg.org/#look-up-a-custom-element-definition will return null if name + local name do not match (in its step 5).

blittle:
Do the tests need to be retried?

src/custom-elements.js
domenic: You can just use `node.localName` instead of `node.tagName.toLowerCase()`. I believe that will be more accurate in Unicode scenarios as well.

joeldenning:
Yeah the tests pass locally for me but timed out in travis. Retrying would probably get them to run.

src/custom-elements.js
joeldenning: 👍  thanks! Just changed this

joeldenning:
Ping

blittle:
I'd love to hear an update about this!

joeldenning:
Ping.

Are there concerns around this pr that are preventing it from being reviewed? Maybe that webcomponents/custom-elements is hesitant about implementing customized built-in elements? Or is it just that since it's a big pull request it is hard to review?

bicknellr:
Sorry, @joeldenning, I owe you an explanation here. I recently ended up overhauling the polyfill so that we could remove most of our dependence on MutationObservers for the sake of performance and timing correctness. Unfortunately, this makes this PR effectively unmergable as there's no common code anymore. I should have paid more attention to what was already happening in the repo and let you know about these changes earlier.

I didn't end up implementing extension for built-ins during that process because, AFAIK, WebKit does not intend to implement it. I'd rather not have people end up stuck on the polyfill because they become dependent on this feature. Also, we won't have to worry about supporting a feature that's already been baked in if it does end up being dropped from the spec.

However, if it does make it through and WebKit decides to implement, then we'll definitely want it here. If you're still interested in implementing this on top of the new polyfill, I'd be glad to help explain how the new polyfill works and point you to the relevant spots. (I'm hoping that the recent updates made the polyfill a bit easier to navigate.) I just can't guarantee it's going to make it in at the moment - you'll have to keep an eye on the spec discussion for that.

joeldenning:
@bicknellr thanks for the explanation. Do you know if/where there is more discussion about Webkit's disagreement with customized built in elements (beyond what is described in [this issue in the CE spec](https://github.com/w3c/webcomponents/issues/509))? I have been following it as best as I can, and as far as I can tell what you said is correct about webkit not planning to implement customized built ins.

About whether this polyfill should implement customized built-ins, though, would you be opposed to including support for customized built-ins if it came with a disclaimer in the documentation? Explaining the current status of the spec (v1 finalized) but that Webkit doesn't plan on implementing a specific part of the spec (customized built-in). Or perhaps as a separate script that is optionally included to support customized built ins?

I think that there is still value in fully implementing the v1 spec, as long as users know what they are getting themselves into. I also think that this polyfill is a great place to educate people on the spec and its browser support, since many developers will come here to learn about custom elements without ever having read the spec or the discussions surrounding it. Once people know that there is a risk that they will be stuck on the polyfill forever if they choose to use customized builtins, I feel that we have given them enough information to make that decision.

All that said, I understand that I'm merely an interested individual that doesn't represent the vast swaths of Polymer users or the large corporations interested in the CE spec 😃 . Interested to hear others' thoughts on whether a polyfill should support a feature that is in the spec but may never be supported by all browsers.

bedeoverend:
@bicknellr If I understand correctly, Chrome, Firefox (written into gecko [source](https://github.com/mozilla/gecko-dev/blob/master/dom/base/CustomElementRegistry.cpp#L856)), and Edge (expressed intent in last [TPAC meeting](https://www.w3.org/2016/09/19-webapps-minutes.html) - potentially could now have changed) all intend to implement built-ins via `is=""` - if the majority of vendors intend to implement, surely this polyfill should implement it also?

I understand wanting to hold off until at least one implementation exists, but if Safari is the only one planning on not implementing, and that was the reason the polyfill held off, the polyfill's functionality would differ from most browsers.

seaneking:
I strongly agree with @joeldenning here - extending built-ins is firmly in the v1 spec, and being actively shipped by multiple vendors. We'd be making a very big assumption that users don't want this polyfilled, because they'd be carrying the polyfill on Webkit, and making that decision for them.

This polyfill should follow the spec, regardless of vendors, and include a disclaimer in the docs that using a certain feature of the spec (extending built-ins) will have performance ramifications on Webkit. Then users are free to make up their own mind.

In practical terms - there are a lot of use-cases that just aren't possible without extending builtins, and having native perf in most vendors with a shim in webkit, vs. hacking something together with mutation observers for _all_ vendors to meet these use-cases, is a compromise that I think a lot of devs would be happy making. 

But, at the end of the day, it should be their choice. A polyfill's job should just be to follow the spec as best as it can and document any edge-cases / caveats. Not to mention that Webkit is the odd one out here, and there is [vocal opposition to their position](https://github.com/w3c/webcomponents/issues/509) from a lot of people. A polyfill shouldn't take sides or dictate usage.

src/custom-elements.js
```diff
@@ -451,6 +459,9 @@ let Deferred;
         const walker = createTreeWalker(root);
         do {
           const node = /** @type {!HTMLElement} */ (walker.currentNode);
```
dominiccooney: FWIW I think this annotation is incorrect; you may have non-HTMLElements in a document, usually SVG.

src/custom-elements.js
```diff
@@ -451,6 +459,9 @@ let Deferred;
...
+          if (node.getAttribute('is')) {
```
dominiccooney: I'm not super familiar with this polyfill, but if it hooks element creation it's more accurate to cache 'is' at creation time. The HTML spec calls this the element's "is value" IIRC.

I guess you probably need to look at attributes for parsed elements.

You may want to keep those values in a side WeakMap of Element -> string. That will avoid the closure type problem you mention below.

src/custom-elements.js
dominiccooney: I don' think this property exists.

src/custom-elements.js
```diff
@@ -562,7 +573,7 @@ let Deferred;
           const node = walker.currentNode;
           if (node[_upgradedProp] && node[_attachedProp]) {
             node[_attachedProp] = false;
-            const definition = this._definitions.get(node.localName);
+            const definition = getDefinition(this._definitions, node);
```
dominiccooney: FWIW the native implementation in Chrome does this differently: once a definition is associated with an element (at creation or upgrade) we store a pointer directly from the element to the definition.

Given that getDefinition is more complicated now, it might be worth revisiting the space/speed tradeoff.

I note that you're storing "is" on the element. If you wanted to save space you could switch the storage space for "is" to point to the full definition once you have it. If you ever happened to need is (I don't think you do; cloning goes back to the attribute value for example) you could retrieve it from the definition.

src/custom-elements.js
joeldenning: Good catch -- I think I confused the ["is value"](https://dom.spec.whatwg.org/#concept-element-is-value) with a property on the dom element. But it seems like the "is value" is just an internal thing not exposed to userland.

src/custom-elements.js
joeldenning: I agree that not looking at the `is` attribute would be ideal, but like you pointed out I don't think it's possible to avoid it for parsed elements. Since (per the spec) updating the `is` attribute after creation should *not* cause an element to be upgraded to a custom element, what I did here was only look at the `is` attribute when the MutationObserver says that a node was *added*, and never look at the `is` attribute after that.

Since this polyfill has been rewritten to rely on monkeypatched methods instead of mutation observers, I'll have to figure out how to get the same functionality without the MutationObserver telling me if it is a newly added node or not.

src/custom-elements.js
joeldenning: 👍  I like that idea. Putting it on the element instead of the looking it up would improve performance. And getting rid of the `is` property will offset the new storage space for the definition.

And regarding the `is` property, just reviewed the spec and looks like I got confused between the ["is value"](https://dom.spec.whatwg.org/#concept-element-is-value) and an actual userland property on the dom element. So you're right, it should be removed from this polyfill

joeldenning:
@dominiccooney thanks for the feedback, will definitely incorporate it!

@bicknellr do you have thoughts on the discussion above about whether to implement customized built-ins for this polyfill? I am more than happy to redo my implementation to work with the monkeypatching approach, if you feel comfortable with that. As discussed above, I think that if we educate users of the polyfill about the potentially shaky status of customized builtins, we've given them the power to choose whether to use customized built-ins or not. I also agree with @bedeoverend that it would be unexpected if the polyfill does not have feature parity with the native implementation of the majority of browsers (Chrome, Firefox, Edge). What are your thoughts?

bicknellr:
@joeldenning Yeah, having a polyfill that doesn't fully implement the spec feels wrong. More so if that spec is implemented by multiple browsers. @dominiccooney mentioned to me that Chrome's built-in extension is behind a flag (maybe that's just because it's not totally there yet?). I think following suit by also putting the polyfill's implementation behind a flag would make sense. That way it's available but still requires opt-in and we can point out the continued debate about the feature where the flag is documented. I think that strikes a reasonable balance between being spec complete and being cautious about allowing users to write code that could potentially be non-portable in the future. WDYT?

joeldenning:
👍  I like that. Maybe something like `customElements.enableCustomizedBuiltIns = true`?

I'll start work on this again (based on the new monkeypatched way) in a new pr.

avinoamr:
@joeldenning How is it looking? Are you still actively working on it? Any estimation on when you'll have something ready?

joeldenning:
Yeah, I have been working on it. I have it partially working, and am just adding tests and fixing bugs. I hope to have it finished and create a pr this upcoming week

avinoamr:
@joeldenning thanks. can't wait :)

PR #43, opened 2017-01-10
WIP: Replace mutation observers with patching.

bicknellr:


src/CustomElementInternals.js
```diff
@@ -0,0 +1,294 @@
...
+
+    const definition = this.localNameToDefinition(element.localName);
+    if (!definition) return;
+
+    definition.constructionStack.push(element);
+
+    const constructor = definition.constructor;
+    try {
+      try {
+        let result = new (constructor)();
```
kevinpschaaf: Do these `try`'s add any perf impact?  Could you just de-FUD by A-B testing with/without?

src/CustomElementInternals.js
```diff
@@ -0,0 +1,294 @@
...
+   */
+  addPatch(listener) {
+    this._patches.push(listener);
+  }
+
+  /**
+   * @param {!Node} node
+   */
+  patchTree(node) {
+    Utilities.walkDeepDescendants(node, node => this.patch(node));
```
kevinpschaaf: This can't be implemented as an additional treewalk of the entire world that typically (on good browsers with no "patches") does nothing.... too expensive.  Should be ok to piggy-back on `upgradeTree`.  Even then it seems a little too cute to have a pluggable patching facility-- do you have other use cases than `innerHTML` and `textContent`?  I'd prefer this be done as cheaply as possible and make sure it's viable before we start generalizing and abstracting.

src/DocumentConstructionObserver.js
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * @param {!Array<!MutationRecord>} mutations
+   */
+  _handleMutations(mutations) {
+    // Once the document's `readyState` is 'interactive' or 'complete', all new
+    // nodes created within that document will be the result of script and
+    // should be handled by patching.
+    const readyState = this._document.readyState;
+    if (readyState === 'interactive' || readyState === 'complete') {
```
kevinpschaaf: Why is the disconnection done here (in `_handleMutations`) and not in a "readystatechange` event handler?  Seems like this would double-walk the first script-created elements appended to the document after 'interactive'...?

src/CustomElementInternals.js
```diff
@@ -0,0 +1,294 @@
...
+  }
+
+  /**
+   * @param {!Node} node
+   */
+  patch(node) {
+    if (node[CESymbols.patched]) return;
```
sorvell: Probably best perf if this can immediately bail on failing a boolean check to a flag that is set if instance patching is needed. I think having a set of patches probably suffers from YAGNI.

src/Patch/Document.js
```diff
@@ -0,0 +1,67 @@
...
+export default function(internals) {
+  Utilities.setPropertyUnchecked(Document.prototype, 'createElement',
+    /**
+     * @this {Document}
+     * @param {string} localName
+     * @return {!Element}
+     */
+    function(localName) {
+      const definition = internals.localNameToDefinition(localName);
+      if (definition) {
```
sorvell: Should only customize elements in the main document.

src/Patch/Document.js
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * @this {Document}
+     * @param {!Node} node
+     * @param {boolean=} deep
+     * @return {!Node}
+     */
+    function(node, deep) {
+      const clone = Native.Document_importNode.call(this, node, deep);
+      internals.patchTree(clone);
+      internals.upgradeTree(clone);
```
sorvell: Again, 2x tree walking is bad.

sorvell: Only the main document should customize.

src/Patch/Document.js
```diff
@@ -0,0 +1,67 @@
...
+     * @return {!Element}
+     */
+    function(namespace, localName) {
+      if (namespace === null || namespace === NS_HTML) {
+        return this.createElement(localName);
+      }
+
+      const result = /** @type {!Element} */
+        (Native.Document_createElementNS.call(this, namespace, localName));
+      internals.patch(result);
```
sorvell: Why must this be patched?

src/Patch/Interface/ParentNode.js
```diff
@@ -0,0 +1,67 @@
...
+    builtIn.prepend.apply(this, nodes);
+
+    for (let i = 0; i < connectedBefore.length; i++) {
+      internals.disconnectTree(connectedBefore[i]);
+    }
+
+    if (Utilities.isConnected(this)) {
+      for (let i = 0; i < nodes.length; i++) {
+        const node = nodes[i];
+        if (node instanceof Element) {
```
sorvell: Don't you have to connect children of added `DocumentFragments`?

src/Patch/Interface/ParentNode.js
```diff
@@ -0,0 +1,67 @@
...
+    builtIn.append.apply(this, nodes);
+
+    for (let i = 0; i < connectedBefore.length; i++) {
+      internals.disconnectTree(connectedBefore[i]);
+    }
+
+    if (Utilities.isConnected(this)) {
+      for (let i = 0; i < nodes.length; i++) {
+        const node = nodes[i];
+        if (node instanceof Element) {
```
sorvell: Again, what about contents of `DocumentFragments`?

src/Patch/Element.js
```diff
@@ -0,0 +1,137 @@
...
+import * as Utilities from '../Utilities';
+
+import PatchParentNode from './Interface/ParentNode';
+import PatchChildNode from './Interface/ChildNode';
+
+/**
+ * @param {!CustomElementInternals} internals
+ */
+export default function(internals) {
+  if (Native.Element_attachShadow) {
```
sorvell: Why is it necessary to patch this?

src/Patch/Element.js
```diff
@@ -0,0 +1,137 @@
...
+  // TODO: Patch instances in browsers without an `Element#innerHTML` descriptor (Early Chrome, IE).
+  if (Native.Element_innerHTML) {
+    Object.defineProperty(Element.prototype, 'innerHTML', {
+      enumerable: Native.Element_innerHTML.enumerable,
+      configurable: true,
+      get: Native.Element_innerHTML.get,
+      set: function(htmlString) {
+        Native.Element_innerHTML.set.call(this, htmlString);
+        internals.patchTree(this);
+        internals.upgradeTree(this);
```
sorvell: Remove double tree walk.

src/Patch/Element.js
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  Utilities.setPropertyUnchecked(Element.prototype, 'setAttribute',
+    /**
+     * @this {Element}
+     * @param {string} name
+     * @param {string} newValue
+     */
+    function(name, newValue) {
+      const oldValue = Native.Element_getAttribute.call(this, name);
```
sorvell: To improve perf for all these patches, suggest checking first if the element is custom and if not do no other work than calling the native function.

src/Patch/Interface/ChildNode.js
```diff
@@ -0,0 +1,107 @@
...
+    builtIn.before.apply(this, nodes);
+
+    for (let i = 0; i < connectedBefore.length; i++) {
+      internals.disconnectTree(connectedBefore[i]);
+    }
+
+    if (Utilities.isConnected(this)) {
+      for (let i = 0; i < nodes.length; i++) {
+        const node = nodes[i];
+        if (node instanceof Element) {
```
sorvell: What about `DocumentFragment` contents?

src/Patch/Interface/ChildNode.js
```diff
@@ -0,0 +1,107 @@
...
+
+    for (let i = 0; i < connectedBefore.length; i++) {
+      internals.disconnectTree(connectedBefore[i]);
+    }
+
+    if (wasConnected) {
+      internals.disconnectTree(this);
+      for (let i = 0; i < nodes.length; i++) {
+        const node = nodes[i];
+        if (node instanceof Element) {
```
sorvell: What about `DocumentFragment` contents?

src/Patch/Node.js
```diff
@@ -0,0 +1,160 @@
...
+      return nativeResult;
+    });
+
+  Utilities.setPropertyUnchecked(Node.prototype, 'appendChild',
+    /**
+     * @this {Node}
+     * @param {!Node} node
+     * @return {!Node}
+     */
+    function(node) {
```
sorvell: Suggest collapsing the implementation of this with `insertBefore` since append is equivalent to `insertBefore` with no ref node.

src/Patch/Node.js
```diff
@@ -0,0 +1,160 @@
...
+  Utilities.setPropertyUnchecked(Node.prototype, 'cloneNode',
+    /**
+     * @this {Node}
+     * @param {boolean=} deep
+     * @return {!Node}
+     */
+    function(deep) {
+      const clone = Native.Node_cloneNode.call(this, deep);
+      internals.patchTree(clone);
+      internals.upgradeTree(clone);
```
sorvell: Avoid double tree walk.

sorvell: Upgrade only if the clone is owned by the main document.

src/Utilities.js
```diff
@@ -0,0 +1,123 @@
...
+  const reserved = reservedTagList.has(localName);
+  const validForm = /^[a-z][.0-9_a-z]*-[\-.0-9_a-z]*$/.test(localName);
+  return !reserved && validForm;
+}
+
+/**
+ * @private
+ * @param {!Node} node
+ */
+export function isConnected(node) {
```
sorvell: Since this is used to determine if lifecycle should be called, it should only return true if `node` is connected to `window.document`.

src/CustomElementInternals.js
sorvell: It looks like `walkDeepDescendants` includes text nodes. Why are text nodes patched?

src/custom-elements.js
```diff
@@ -8,792 +8,47 @@
...
+  PatchNode(internals);
+  PatchElement(internals);
+
+  /** @type {!CustomElementRegistry} */
+  const customElements = new CustomElementRegistry(internals);
+
+  // If `customElements.documentReady` exists and is a Promise, prevent calls to
+  // define from causing upgrades until `documentReady` has resolved.
+  if (priorCustomElements) {
+    const documentReady = priorCustomElements['documentReady'];
```
sorvell: I think we should change the design of this hook to be a function that can return true or a promise. The reason is that this state is dynamic (lazy loading) and we want it to be synchronous so that upgrades happen synchronously when possible.

src/Utilities.js
```diff
@@ -0,0 +1,123 @@
...
+    node = node.parentNode || (node.nodeType === Node.DOCUMENT_FRAGMENT_NODE ? node.host : undefined);
+  }
+  return node instanceof Document;
+}
+
+/**
+ * @param {!Node} root
+ * @param {!function(!Element)} callback
+ */
+export function walkDeepDescendantElements(root, callback) {
```
kevinpschaaf: I'm not sure the design of "capture tree being added/removed into flat list before running callbacks, then call callbacks on the captured list" that was added in this PR is a good tradeoff of performance for correctness, at least as implemented.

In the V1/current V1 polyfills, the MO record just gave the parent node that was added removed, then it was treewalked and callbacks called during the walk.  This meant that if a connected callback of a parent removed children, those children would not have their connected callbacks called (which is technically incorrect, although they would just be re-disconnected immediately after).

Pre-capturing the list (which is more expensive since you have the added cost of constructing the array then iterating it) seems halfsies -- in the above example it ensures that connected callback _is_ called, but it seems like it will come _after_ the disconnected callback is run, which would be _more_ incorrect since the callbacks come out of order.  To correct this it seems like you'd need an even more correct (and expensive) callback queue, to ensure the callbacks are run in the correct order.  Like we've discussed before I'm ok with exploring that, but in the name of performance and time, it seems like the previous behavior (less correct but not broken, and faster) is a better tradeoff for the MVP changes we need right now.

src/Patch/Document.js
```diff
@@ -0,0 +1,67 @@
...
+import CustomElementInternals from '../CustomElementInternals';
+import * as Utilities from '../Utilities';
+
+import PatchParentNode from './Interface/ParentNode';
+
+/**
+ * @param {!CustomElementInternals} internals
+ */
+export default function(internals) {
+  Utilities.setPropertyUnchecked(Document.prototype, 'createElement',
```
kevinpschaaf: As discussed, these three methods need to check for `this === window.document` before customizing.

src/Utilities.js
```diff
@@ -0,0 +1,123 @@
...
+      }
+
+      if (!next || context === root) return;
+
+      walker.currentNode = next;
+      walker.previousNode();
+      continue;
+    }
+
+    const shadowRoot = element[CESymbols.shadowRoot];
```
kevinpschaaf: Why not just `element.shadowRoot`? Why patch attachShadow and store it off on a different name?

src/CustomElementInternals.js
```diff
@@ -0,0 +1,294 @@
...
+    }
+  }
+
+  /**
+   * @param {!Element} element
+   */
+  connectedCallback(element) {
+    if (element[CESymbols.state] === CEState.custom) {
+      const definition = element[CESymbols.definition];
+      if (definition && definition.connectedCallback) {
```
bicknellr: `definition` is always defined if the element is custom

src/DocumentConstructionObserver.js
bicknellr: (use `takeRecords`?)

src/Patch/Document.js
```diff
@@ -0,0 +1,67 @@
...
+  Utilities.setPropertyUnchecked(Document.prototype, 'createElementNS',
+    /**
+     * @this {Document}
+     * @param {?string} namespace
+     * @param {string} localName
+     * @return {!Element}
+     */
+    function(namespace, localName) {
+      if (namespace === null || namespace === NS_HTML) {
+        return this.createElement(localName);
```
bicknellr: This should use `Native.Document...` instead of `this` to call `createElement`.

src/Patch/Element.js
bicknellr: (from offline) This is for closed shadow roots.

src/Utilities.js
```diff
@@ -0,0 +1,123 @@
...
+  return !reserved && validForm;
+}
+
+/**
+ * @private
+ * @param {!Node} node
+ */
+export function isConnected(node) {
+  // Use `Node#isConnected`, if defined.
+  const nativeValue = node.isConnected;
```
bicknellr: (from offline) Native `isConnected` returns true if the element has a deep root that is *any* document, but `{dis}connectedCallback` should only be called if the element moves in and out of (the main document and html import documents of connected links (?)). Relevant documents need to be marked when crawling in `upgradeTree`.

src/Utilities.js
```diff
@@ -0,0 +1,123 @@
...
+      }
+    }
+  } while (walker.nextNode());
+}
+
+/**
+ * @param {!Node} root
+ * @param {!function(!Node)} callback
+ */
+export function walkDeepDescendants(root, callback) {
```
bicknellr: Remove this because text nodes aren't being patched.

src/CustomElementInternalSymbols.js
```diff
@@ -0,0 +1,7 @@
+const randomString = Math.random().toString(32).substring(2);
+
+export const state = '_CE_state_' + randomString;
+export const definition = '_CE_definition_' + randomString;
+export const shadowRoot = '_CE_shadowRoot_' + randomString;
+export const documentLoadHandled = '_CE_documentLoadHandled_' + randomString;
+export const patched = '_CE_patched_' + randomString;
```
bicknellr: This should be fixed now. I added externs for these names so that closure doesn't mangle them and I just reference them directly, rather than with `[]`.

src/CustomElementInternals.js
```diff
@@ -0,0 +1,294 @@
...
+  upgradeTree(root, visitedImports) {
+    if (!visitedImports) {
+      visitedImports = new Set();
+    }
+
+    const elements = [];
+
+    const gatherElements = element => {
+      if (element.localName === 'link' && element.getAttribute('rel') === 'import') {
+        if (visitedImports.has(element)) return;
```
bicknellr: This should be fixed.

src/Utilities.js
bicknellr: `isConnected` is true as long as a node's shadow-including root is a document but I've also updated this to bail out early if we hit an import's root node (a document in native, an element which I've marked with a flag in the polyfill).

src/CustomElementRegistry.js
```diff
@@ -0,0 +1,163 @@
...
+    if (!Utilities.isValidCustomElementName(localName)) {
+      throw new SyntaxError(`The element name '${localName}' is not valid.`);
+    }
+
+    if (this._internals.localNameToDefinition(localName)) {
+      throw new Error(`A custom element with name '${localName}' has already been defined.`);
+    }
+
+    if (this._elementDefinitionIsRunning) {
+      throw new Error('A custom element is already being defined.');
```
bicknellr: yup

src/Patch/Document.js
bicknellr: This should be fixed. I mark documents that are meant to be associated with the registry with a flag and only customize elements if they're created in the context of one of those documents.

src/Patch/Element.js
bicknellr: These should be fixed in both of each `setAttribute` and `removeAttribute`.

src/Utilities.js
```diff
@@ -0,0 +1,123 @@
...
+  // Use `Node#isConnected`, if defined.
+  const nativeValue = node.isConnected;
+  if (nativeValue !== undefined) {
+    return nativeValue;
+  }
+
+  while (node && !(node instanceof Document)) {
+    node = node.parentNode || (node.nodeType === Node.DOCUMENT_FRAGMENT_NODE ? node.host : undefined);
+  }
+  return node instanceof Document;
```
bicknellr: [Any document apparently works](https://dom.spec.whatwg.org/#connected), which is kind of weird since you get `connectedCallback` when connecting a custom element to any script-created document.

src/Patch/Document.js
bicknellr: This should be fixed now.

src/Utilities.js
bicknellr: Ignore the comment above, `{,dis}connectedCallback` always gets called when `isConnected` changes.

src/CustomElementInternals.js
bicknellr: It only patches elements now, I still need to merge `patchTree` into `upgradeTree`.

src/CustomElementInternals.js
```diff
@@ -0,0 +1,294 @@
...
+        if (element.import instanceof Node) {
+          Utilities.walkDeepDescendantElements(element.import, gatherElements);
+        } else {
+          element.addEventListener('load', () => {
+            const doc = element.import;
+
+            if (doc[CESymbols.documentLoadHandled]) return;
+            doc[CESymbols.documentLoadHandled] = true;
+
+            this.upgradeTree(document);
```
bicknellr: This should be fixed.

src/CustomElementInternals.js
bicknellr: I've added a flag that gets flipped on the first call to `addPatch`, so both `patchTree` and `patch` should bail if there's nothing to do. I'd prefer not to have to put all the patches in one place; ideally they would be grouped with all the other (non-instance) patches that are being applied to a particular prototype. (`innerHTML` defined in the same place as other patches for Element and `textContent` with Node, if it ends up being necessary.)

src/Patch/Document.js
bicknellr: Probably not necessary? I'm not sure `innerHTML` is even well defined enough to know what should happen when `innerHTML` is read/written on elements in other namespaces. However, given that `innerHTML` is on the Element prototype - as opposed to HTMLElement - I think patching this would be more correct.

src/Patch/Document.js
bicknellr: Upgrades should only happen in the context of documents with an associated registry now. Still need to avoid the double walk.

src/Patch/Node.js
bicknellr: Upgrades should only happen in the context of documents with an associated registry now. Still need to avoid the double walk.

src/Utilities.js
```diff
@@ -0,0 +1,123 @@
...
+    callback(element);
+
+    // Ignore descendants of link elements to prevent attempting to traverse
+    // elements created by the HTML Imports polyfill. When upgrading a tree,
+    // `CustomElementInternals#upgradeTree` by walks the `import` property of
+    // import links, which will point to either the true imported document
+    // (native) or the descendant containing the elements created to emulate the
+    // imported document (polyfill).
+    if (element.localName === 'link') {
+      let context = element;
```
bicknellr: I think it should be `walker.currentNode = element.lastChild || element` (in case the link has no children). But yes, it looks like this whole section is unnecessary.

src/Utilities.js
bicknellr: This is also for closed shadow roots.

src/Utilities.js
bicknellr: fixed

src/DocumentConstructionObserver.js
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * @type {MutationObserver|undefined}
+     */
+    this._observer = undefined;
+
+
+    // Simulate tree construction for all currently accessible nodes in the
+    // document.
+    this._internals.patchTree(this._document);
+    this._internals.upgradeTree(this._document);
```
bicknellr: This should be fixed now, `upgradeTree` now attempts to patch elements before upgrading.

src/DocumentConstructionObserver.js
```diff
@@ -0,0 +1,61 @@
...
+    if (readyState === 'interactive' || readyState === 'complete') {
+      this._observer.disconnect();
+    }
+
+    for (let i = 0; i < mutations.length; i++) {
+      const addedNodes = mutations[i].addedNodes;
+      for (let j = 0; j < addedNodes.length; j++) {
+        const node = addedNodes[j];
+        this._internals.patchTree(node);
+        this._internals.upgradeTree(node);
```
bicknellr: (also fixed)

src/Patch/Document.js
bicknellr: (double walk fixed)

src/Patch/Element.js
bicknellr: (fixed)

src/Patch/Node.js
bicknellr: (double walk fixed)

src/CustomElementInternals.js
bicknellr: Patching now happens during upgrades.

src/CustomElementInternals.js
```diff
@@ -0,0 +1,294 @@
...
+    if (Utilities.isConnected(element)) {
+      this.connectedCallback(element);
+    }
+  }
+
+  /**
+   * @param {!Element} element
+   */
+  connectedCallback(element) {
+    if (element[CESymbols.state] === CEState.custom) {
```
bicknellr: Fixed; I'm not a fan of this though.

src/custom-elements.js
bicknellr: I put this in [a separate PR](https://github.com/webcomponents/custom-elements/pull/54) based off of this branch, just to make sure the API is what you're expecting.

src/Utilities.js
```diff
@@ -55,26 +54,16 @@ export function walkDeepDescendantElements(root, callback) {
...
-      while (!next && context !== root) {
-        context = context.parentNode;
-        next = context.nextSibling;
-      }
-
-      if (!next || context === root) return;
-
-      walker.currentNode = next;
-      walker.previousNode();
+      walker.currentNode = element.lastChild || element;
```
kevinpschaaf: Whoops, looks like this didn't work.  It'll still plunge into children of the link's last child, which is not what we want.  Sorry!!

This seems simpler than the old way:

```js
while (element.lastChild) {
  element = element.lastChild;
}
walker.currentNode = element;
```

src/custom-elements.js
```diff
@@ -28,18 +28,21 @@ if (!priorCustomElements || priorCustomElements['forcePolyfill']) {
...
   /** @type {!CustomElementRegistry} */
   const customElements = new CustomElementRegistry(internals);
 
   // If `customElements.documentReady` exists and is a Promise, prevent calls to
   // define from causing upgrades until `documentReady` has resolved.
-  if (priorCustomElements) {
+  if (priorCustomElements && priorCustomElements['documentReady']) {
     const documentReady = priorCustomElements['documentReady'];
-    if (documentReady && documentReady.then instanceof Function) {
+    if (documentReady.then instanceof Function) {
```
kevinpschaaf: Hopefully you just haven't gotten to it yet, but just to capture what we discussed last week, the API & logic for `documentReady` needs to match the `master` version of the v1 polyfill to meet backwards-compatible timing guarantees:
* `documentReady` can't be a promise because it needs to be able to return synchronously if it is ok to upgrade elements, so that `customElements.define('x-foo', XFoo); assert.true(document.querySelector('x-foo') instanceof XFoo);` passes if there are no imports in flight.
* `customElements.define` needs to _call_ `documentReady` (should debounce so that only one upgrade treewalk happens if multiple `defines` occur when imports are in flight)
* `whenDefined` promises should resolve only once any treewalk pending on `domcumentReady` are resolved.

Please see the logic here: https://github.com/webcomponents/custom-elements/blob/master/src/custom-elements.js#L351-L366

src/CustomElementRegistry.js
```diff
@@ -101,7 +100,7 @@ export default class CustomElementRegistry {
     this._internals.setDefinition(localName, definition);
 
     if (this._upgradeOnDefine) {
-      this._internals.upgradeTree(document);
+      this._internals.patchAndUpgradeTree(document);
```
kevinpschaaf: See https://github.com/webcomponents/custom-elements/pull/43/files/d3c10972dee961702ab5248dc871416207d339c0..2f22e61d91e1f2ef8be6c5bbb53d6a702ecb819b#r97923523; this is the work that needs to be gated on a debounced _call_ to `documentReady` (rather than checking a flag). The function call as opposed to a flag is important so that imports polyfill can synchronously check for newly added imports.

src/custom-elements.js
bicknellr: Sorry I didn't notice this comment earlier, I put this change in a PR #54, PTAL. I think the API matches what you've described:
- If the callback is defined, `customElements.define` will call the callback with a function; this function can be used to trigger the document walk.
```javascript
window.customElements = {
  flushCallback: function(doFlush) {
    otherThing.doFlush = doFlush;
  },
};
// ... custom elements polyfill runs here ...
customElements.define('element-a', class extends HTMLElement {});
// Synchronously during the above call, `flushCallback` is called with a function.
```
- If multiple calls to `customElements.define` occur, only the first will cause the callback to be called (with the function).
```javascript
// .. continuing from above.
customElements.define('element-b', class extends HTMLElement {});
//`flushCallback` is *not* called.
```
- If the function *given* to the callback is called to trigger a document walk, the *next* `customElements.define` call will again call the callback (giving the document-walk-triggering function).
```javascript
// .. continuing from above.
otherThing.doFlush();
// ^^ This causes the document to be walked and promises given by `whenDefined`
// for 'element-a' and 'element-b' to be resolved.
customElements.define('element-c', class extends HTMLElement {});
// `flushCallback` is called with the upgrade-triggering function again.
```
I don't think we discussed how to disable this, so the way I handle this is to provide a function called `customElements.polyfillSetFlushCallback` that you can use to add / change / remove the callback. Giving it a function starts using that function as the callback; giving it `undefined` automatically flushes anything that was delayed and stops using the last callback. Similar to `forcePolyfill`, if a property named `flushCallback` exists on a global object called `customElements` when the polyfill starts up, it will use this function initially.
```javascript
// .. continuing from above.
customElements.polyfillSetFlushCallback(undefined);
// ^^ The document is walked and promises for 'element-c' are resolved.
// Later calls to `customElements.define` do *not* call the callback and, instead,
// they synchronously walk the document and resolve `whenDefined` promises.
```

bicknellr: FYI, DOM operations that would cause an element to be constructed or a tree to be crawled for upgrades (`importNode`, `createElement`, setting `innerHTML`) that occur when this delayed-document-walk thing is happening will still happen. The definition itself hasn't been blocked, just the associated full-document walk and `whenDefined` promise resolution.

src/Utilities.js
bicknellr: Good catch; should be fixed now.

src/Utilities.js
```diff
@@ -0,0 +1,129 @@
...
+  // Use `Node#isConnected`, if defined.
+  const nativeValue = node.isConnected;
+  if (nativeValue !== undefined) {
+    return nativeValue;
+  }
+
+  /** @type {?Node|undefined} */
+  let current = node;
+  while (current && !(current.__CE_isImportDocument || current instanceof Document)) {
+    current = current.parentNode || (current instanceof ShadowRoot ? current.host : undefined);
```
valdrinkoshi: `instanceof ShadowRoot` fails on browsers w/o `window.ShadowRoot` (e.g. IE11)

bower.json
```diff
@@ -16,7 +16,10 @@
...
   "ignore": [],
   "devDependencies": {
-    "web-component-tester": "^4.0.1",
-    "webcomponentsjs": "webcomponents/webcomponentsjs#^0.7.22"
+    "web-component-tester": "6.0.0-prerelease.3",
+    "webcomponentsjs": "webcomponents/webcomponentsjs#gulp-all-the-things",
+    "webcomponents-platform": "webcomponents/platform#master",
+    "es6-promise": "stefanpenner/es6-promise#^4.0.0",
+    "html-imports": "webcomponents/html-imports#remove-timeout"
   }
```
valdrinkoshi: Wrong dependency here, should be master

bicknellr: Thanks, updated!

PR #52, opened 2017-01-25
Fix #44

### LINKED ISSUE #44: `attributeChangedCallback` is called twice when the attribute is changed
Hello,

On the `custom-element` polyfill, the function `attributeChangedCallback` is called twice when an attribute, listed in the `observedAttributes()`, is changed.

```html
<my-component id="compo" foo="bar"></my-component>
<button id="myBtn">Change value</button>
```

```js
class MyComponent extends HTMLElement {
  constructor() {
    super();
  }
  
  static get observedAttributes() { return ['foo']; }
  
  attributeChangedCallback(attrName, oldVal, newVal) {
    console.log('[my component] attribute', attrName, 'changed from', oldVal, 'to', newVal);
  }
}

window.customElements.define('my-component', MyComponent);
```

The logs are as follows:

```
[my component] attribute foo changed from null to bar
click on the button to change the attribute value
[my component] attribute foo changed from bar to baz
[my component] attribute foo changed from bar to baz
```

As you can see, the line `[my component] attribute foo changed from bar to baz` is called twice.

Here is a JsFiddle to test that example: https://jsfiddle.net/czo1355c/

The issue is encountered on a Chrome v53, since starting with v54, `customElements` is natively supported by Chrome (and the function `attributeChangedCallback` is only called once).
Tested against version `1.0.0-alpha.3` of custom-element library.

Thanks

chiefcll:
The change attribute is updating the element to the value of baz. The newValue will be baz and it was being compared to the oldValue or bar, but since it was updated, you should compare the newValue to the value being passed in, which is the value we want.

### DIFF
--- a/src/custom-elements.js
+++ b/src/custom-elements.js
@@ -777,7 +777,7 @@ let Deferred;
       const attributeChangedCallback = definition.attributeChangedCallback;
       if (attributeChangedCallback && observedAttributes.indexOf(name) >= 0) {
         const newValue = element.getAttribute(name);
-        if (newValue !== oldValue) {
+        if (newValue !== value) {
           attributeChangedCallback.call(element, name, oldValue, newValue, null);
         }
       }


bicknellr:
Thanks for submitting this PR; this should be fixed now though. Let me know if you run into any issues.

PR #54, opened 2017-01-26
Allow document walks in `customElements.define` to be delayed.

bicknellr:
.. Is this the expected API for this feature?

### DIFF
--- a/src/CustomElementRegistry.js
+++ b/src/CustomElementRegistry.js
@@ -148,8 +148,8 @@ export default class CustomElementRegistry {
   }
 
   /**
-   * Calling `polyfillSetFlushCallback` with a function will set that function
-   * as the registry's 'flush callback'.
+   * Setting `polyfillFlushCallback` to a function will set that function as the
+   * registry's 'flush callback'.
    *
    * If the flush callback is `undefined` (the default), calls to `define` cause
    * synchronous full-document walks to find and upgrade elements and resolve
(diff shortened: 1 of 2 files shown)


src/CustomElementRegistry.js
```diff
@@ -99,14 +111,98 @@ export default class CustomElementRegistry {
...
+      if (!this._flushScheduled) {
+        this._flushScheduled = true;
+        this._flushCallback(() => this._doFlush());
+      }
+    }
+  }
+
+  _doFlush() {
+    // If no new definitions were defined, don't attempt to flush.
+    if (this._flushScheduled === false) return;
```
kevinpschaaf: This case should not be possible, yeah?

src/CustomElementRegistry.js
```diff
@@ -99,14 +111,98 @@ export default class CustomElementRegistry {
...
+   *
+   * // Setting the flush callback to `undefined` walks the document for
+   * // upgrades and resolves any pending `whenDefined` promises, if any calls
+   * // to `define` were made since the last flush.
+   * customElements.polyfillSetFlushCallback(undefined);
+   * ```
+   *
+   * @param {Function} flushCallback
+   */
+  polyfillSetFlushCallback(flushCallback) {
```
kevinpschaaf: Let's make this a get/set `polyfillFlushCallback` accessor, and make it match the `priorCustomElements` interface.

src/CustomElementRegistry.js
bicknellr: If you hold on to one of the functions and call it multiple times, possibly.

src/CustomElementRegistry.js
bicknellr: Fixed.

PR #57, opened 2017-02-09
Guard patches for `ChildNode` and `ParentNode`.

bicknellr:


jab:
It looks like I just hit issue #103 while testing my Polymer 2 site in IE11, and according to https://github.com/webcomponents/custom-elements/issues/103#issuecomment-310524812 this patch fixes it – since there was no mention of that here, I've just added one. :) When someone with the commit bit has a chance to take a look, I'll be interested to see if this fixes the issue. Thanks!

bicknellr:
This PR needs tests that explicitly check that the existence of these methods matches before and after the polyfill is applied.

bicknellr:
Closing: #87 will fix this and also includes tests and other fixes for these methods.

PR #60, opened 2017-02-14
Move bower install into a different npm script

### LINKED ISSUE #59: Install script shouldn't call bower install
https://github.com/webcomponents/custom-elements/blob/8b1c47ff790d51678f75c7016b61ad9655156cc0/package.json#L19

This script will run when anyone installs `webcomponents/custom-elements` via npm. This is not what you want, and makes it so that installation fails unless the person had bower installed globally. For example, in CI. 

To work around this, I had to have Travis install bower globally.

I would suggest putting this command under another script so you do `npm run bower`. I know this is less convenient 😟, but npm doesn't have any hooks that only run when working locally.

matthewp:
This moves the `bower install` command into a different npm script. The
"install" script is ran any time someone npm installs this package.
Since this is only used for local development, moving this into a `bower
install` command.

Closes #59

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ addons:
     - google-chrome-stable
 before_script:
 - export PATH=$PWD/node_modules/.bin:$PATH
+- npm run bower-install
 script:
 - npm run build
 - xvfb-run wct
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "url": "https://github.com/webcomponents/custom-elements/issues"
   },
   "scripts": {
-    "install": "$(npm bin)/bower install",
+    "bower-install": "$(npm bin)/bower install",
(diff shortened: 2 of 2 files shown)


matthewp:
Please merge this soon. This repository cannot be used with npm until this is merged. Thank you.

smalluban:
@matthewp Travis config should be updated to run `npm run bower-install`, because now bower is not executed.

in `.travis.yml`:
```
before_script:
- export PATH=$PWD/node_modules/.bin:$PATH
- npm run bower-install // add this line
```

matthewp:
@smalluban Thanks! I've updated the PR with that change.

bicknellr:
Thanks for pointing this out! My npm-fu could use some work.

matthewp:
No problem, thanks for the merge!

Issue #85, opened 2017-03-24
attributeChangedCallback must be called after every setAttribute


cronon:
Currently this callback is called only when the value of the attribute has changed. Spec says to call it even when the value hasn't changed.
https://www.w3.org/TR/custom-elements/#concept-element-attributes-change
https://bugs.chromium.org/p/chromium/issues/detail?id=676247

PR #86, opened 2017-03-24
call attrChangedCb even if value hasn't changed

cronon:
Spec says to call it even when the value hasn't changed.
https://www.w3.org/TR/custom-elements/#concept-element-attributes-change
https://bugs.chromium.org/p/chromium/issues/detail?id=676247

resolves https://github.com/webcomponents/custom-elements/issues/85


bicknellr:
Hi @cronon, thanks for submitting this! It looks like you need to [sign the CLA](https://cla.developers.google.com).

cronon:
Hi @bicknellr, I thought I've already signed it. How can I check it?

bicknellr:
I think if you click the button at the bottom of the page ("Manage your agreements") it should show you which agreements you've signed, if any.

cronon:
Yes, there is agreement signed. So, how do you check whether contributors have it?

bicknellr:
Found it, thanks! We've got a tool that lets us look up people by their github username / email to check if they've signed it; I'm not sure if it's accessible externally though.

bicknellr:
One more thing, should this change also be made to `setAttributeNS` (directly below)?

cronon:
Right, I forgot this method even exists :)

PR #87, opened 2017-03-29
Fix document fragment interfaces

bicknellr:
This still needs some work. I'm not sure if it was a good idea to merge `before`/`after` and `prepend`/`append`. Also, it looks like I forgot to block the tests in browsers where these functions are not defined.


bicknellr:
@azakus this is the PR that happens to also cover all of #57. I don't think my original comment about running the tests in browsers that didn't support these methods in the first place is still applicable but the patches for similar methods are still merged.

PR #88, opened 2017-04-01
Implementing customized builtin elements.

internettrans:
See https://github.com/webcomponents/custom-elements/pull/42 for previous discussion about customized builtin elements.

Note that I was unable to run some of the html tests locally (I think related to [this issue in wct](https://github.com/Polymer/web-component-tester/issues/416)), and because of that was also unable to add more tests to several files. Let me know what you want me to do about that. Also note that I'm not very well versed in Closure types and welcome any help with that.

Fixes https://github.com/webcomponents/custom-elements/issues/13


src/CustomElementInternals.js
```diff
@@ -4,7 +4,7 @@ import CEState from './CustomElementState.js';
 export default class CustomElementInternals {
   constructor() {
     /** @type {!Map<string, !CustomElementDefinition>} */
-    this._localNameToDefinition = new Map();
+    this._nameToDefinition = new Map();
```
joeldenning: For autonomous custom elements, the custom element name is the same as `localName`. But for customized builtin elements, the `localName` is different than the name of the custom element. I have changed this Map so that it looks thing up by `name` instead of `localName`, so that it works with both autonomous and customized builtin elements.

src/CustomElementInternals.js
```diff
@@ -42,6 +42,20 @@ export default class CustomElementInternals {
   }
 
   /**
+   * @param {!Node} node
+   * @return {!CustomElementDefinition|null}
+   */
+  nodeToDefinition(node) {
```
joeldenning: Getting the CE definition for a DOM element used to be as straightforward as looking things up by `localName`. But now it's a bit more complicated since you have to deal with customized builtins where the localName is not the same as the name.

src/CustomElementInternals.js
```diff
@@ -42,6 +42,20 @@ export default class CustomElementInternals {
...
+    const name = typeof node.__CE_is === 'string' ? node.__CE_is : node.localName;
```
joeldenning: `__CE_is` is a new variable I have created on nodes that are customized builtin elements.

src/CustomElementInternals.js
```diff
@@ -42,6 +42,20 @@ export default class CustomElementInternals {
...
+    const definition = this.nameToDefinition(name);
+    if (definition && definition.localName === node.localName) {
```
joeldenning: It is possible that someone tries to apply a customized builtin element (via `is="my-builtin"`) to a dom element that the customized builtin was never meant to extend. In that case, we should not upgrade the element to be the customized builtin.

This is why there is the `definition.localName === node.localName` check.

src/CustomElementInternals.js
```diff
@@ -208,6 +222,11 @@ export default class CustomElementInternals {
             this.patchAndUpgradeTree(importNode, visitedImports);
           });
         }
+      } else if (element.hasAttribute('is') || element.is) {
```
joeldenning: The `is=""` attribute should only be looked at when an element is being created, but never again. This function is run when `innerHTML` occurs, so it is when elements are newly created. So this is the only place we respect what is in the `is` attribute.

src/CustomElementInternals.js
```diff
@@ -208,6 +222,11 @@ export default class CustomElementInternals {
...
+        element.__CE_is = element.getAttribute('is') || element.is;
+        element.removeAttribute('is');
```
joeldenning: We remove the is attribute so that browsers don't try to do anything native with it.

src/CustomElementRegistry.js
```diff
@@ -47,7 +47,7 @@ export default class CustomElementRegistry {
      * @private
      * @type {!Array<string>}
      */
-    this._unflushedLocalNames = [];
+    this._unflushedNames = [];
```
joeldenning: Similar to `[localNameToDefinition](https://github.com/webcomponents/custom-elements/pull/88/files#diff-4b635b4d15236896a8e79c481a320662R32)`, we now need to keep track of things based on the custom element name, not on the local name. This is because customized builtins have a different localName than a CE name.

src/CustomElementRegistry.js
```diff
@@ -57,20 +57,39 @@ export default class CustomElementRegistry {
...
-    if (this._internals.localNameToDefinition(localName)) {
-      throw new Error(`A custom element with name '${localName}' has already been defined.`);
+    if (this._internals.nameToDefinition(name)) {
+      throw new Error(`A custom element with name '${name}' has already been defined.`);
+    }
+
+    let localName = name;
+
+    if (options && options.extends) {
+      if (!this.enableCustomizedBuiltins) {
```
joeldenning: This is what puts customized builtins behind a flag.

src/CustomElementRegistry.js
```diff
@@ -57,20 +57,39 @@ export default class CustomElementRegistry {
...
 
-    if (this._internals.localNameToDefinition(localName)) {
-      throw new Error(`A custom element with name '${localName}' has already been defined.`);
+    if (this._internals.nameToDefinition(name)) {
+      throw new Error(`A custom element with name '${name}' has already been defined.`);
+    }
+
+    let localName = name;
+
+    if (options && options.extends) {
```
joeldenning: See `6.` and `7.` in [the spec for define](https://html.spec.whatwg.org/multipage/scripting.html#customelementregistry)

src/Patch/HTMLElementSubclasses.js
```diff
@@ -0,0 +1,69 @@
+import Native from './Native.js';
+import CustomElementInternals from '../CustomElementInternals.js';
+import CEState from '../CustomElementState.js';
+import AlreadyConstructedMarker from '../AlreadyConstructedMarker.js';
+
+/**
+ * @param {!CustomElementInternals} internals
+ */
+export default function(internals) {
```
joeldenning: Browsers currently throw errors if you try to subclass native DOM elements. So we patch each of the DOM element constructors so that they don't throw errors.

src/Patch/Native.js
```diff
@@ -28,4 +28,71 @@ export default {
   HTMLElement: window.HTMLElement,
   HTMLElement_innerHTML: Object.getOwnPropertyDescriptor(window.HTMLElement.prototype, 'innerHTML'),
   HTMLElement_insertAdjacentElement: window.HTMLElement.prototype['insertAdjacentElement'],
+  HTMLElement_subclasses: {
```
joeldenning: This is used when patching all the native elements.

src/custom-elements.js
```diff
@@ -35,6 +37,7 @@ if (!priorCustomElements ||
 
   /** @type {!CustomElementRegistry} */
   const customElements = new CustomElementRegistry(internals);
+  customElements.enableCustomizedBuiltins = priorCustomElements.enableCustomizedBuiltins;
```
joeldenning: If somebody set `enableCustomizedBuiltins` before the polyfill executes, we need to respect that.

src/CustomElementInternals.js
```diff
@@ -208,6 +222,11 @@ export default class CustomElementInternals {
...
+        delete element.is;
+        elements.push(element);
```
valdrinkoshi: You could limit the calls to `hasAttribute, getAttribute, removeAttribute` to have better perf. Also, avoid `delete element.is` as it's slow and all we need here is to unset it.

e.g.
```js
} else {
  // if getAttribute returns an empty string consider it null.
  const is = element.is || element.getAttribute('is') || null;
  if (is) {
    element.__CE_is = is;
    element.is = null;
    element.removeAttribute('is');
  }
  elements.push(element);
}

File: src/CustomElementInternals.js (hunk 3 above)
Comment (valdrinkoshi) [inline src/CustomElementInternals.js]: nit: `const name = node.__CE_is || node.localName;`

**Review by bicknellr — COMMENTED** [review]
Nice work! I left a handful of comments, PTAL.

I think `__CE_is` *might* be unnecessary since `__CE_definition` will contain `name` and `localName` in it. Although, maybe there are other places where we need to know the element's `is` value that I'm not thinking about?

File: src/Patch/HTMLElementSubclasses.js [hunk 14]
```diff
@@ -0,0 +1,69 @@
...
+        if (!definition) {
+          throw new Error('The custom element being constructed was not registered with `customElements`.');
+        }
+
+        const constructionStack = definition.constructionStack;
+
+        if (constructionStack.length === 0) {
+          let element;
+          if (definition.name !== definition.localName) {
+            element = Native.Document_createElement.call(document, definition.localName, {is: definition.name});
```
joeldenning: Good point. What I was wanting was the `is` attribute to be set for the customized builtin element. But really I should just do `setAttribute('is', definition.name)`

I have updated this, and also added a test to verify that the `is` attribute gets added when you construct a customized builtin with `new XFoo()`

src/Patch/HTMLElementSubclasses.js
```diff
@@ -0,0 +1,69 @@
...
+          } else {
+            element = Native.Document_createElement.call(document, definition.localName);
+          }
+          Object.setPrototypeOf(element, constructor.prototype);
+          element.__CE_state = CEState.custom;
+          element.__CE_definition = definition;
+          internals.patch(element);
+          return element;
```
joeldenning: I fixed the warning by changing it to `/** @type {!HTMLElement} */ (element)`, similar to [line 61](https://github.com/webcomponents/custom-elements/pull/88/files#diff-b7a7888f33daa329334b5f8beb28457dR61).

The warning was that `Native.Document_createElement` returns a `Node{....}` instead of an `HTMLElement` like is expected.

src/custom-elements.js
joeldenning: I updated this to work with Closure's `ADVANCED` mode. It was mangling `enableCustomizedBuiltins`, but I fixed that.

This is why the tests were failing in the build. Shouldn't happen anymore

tests/index.html
```diff
@@ -12,6 +12,7 @@
 <meta charset="utf-8">
 <script>
   (window.customElements = window.customElements || {}).forcePolyfill = true;
+  window.customElements.enableCustomizedBuiltins = true;
```
joeldenning: https://github.com/webcomponents/custom-elements/pull/88/files#diff-2a35783ddd795619bc384b6c770fe627R90 is a test that does just that

seaneking:
Bump. Any update on this? Very keen to try it out across several elements on [Simpla](https://www.simpla.io) and [Simple UI](https://webcomponents.org/collection/SimpleElements/simple-elements)

joeldenning:
@seaneking I've updated the pr yesterday to address all the comments, look forward to seeing what other thoughts @bicknellr has

avinoamr:
@bicknellr Can we merge this in? Is there anything else missing here?

bicknellr:
Sorry about the delay! We're still full steam ahead on hybrid porting stuff over in Polymer element land so I won't be able to give this a thorough review for a little bit. However, we're really close and I definitely want to see this landed so, once that's all settled, I'll be able to give this PR more attention.

joeldenning:
ping

seaneking:
Pong?

Is there anything people from the community can do to help this get merged? It's holding up a lot of us from migrating existing type-extended elements to WC v1 / Polymer 2.

notwaldorf:
I am a little bit worried about this PR. My understanding is that is= isn't implemented on _any_ browser right now, not even Chrome, which means you would force the polyfill in Chrome for your entire app, which pre-this PR it wouldn't be necessary. Secondly, Apple has been pretty clear they don't want to implement is=, which again means you would always have to force the polyfill. Forcing the polyfill makes me sad, because it's slower and more expensive than the native platforms :(

In Polymer 2 we've just worked with the assumption that is= is not available, and that you have to wrap all your native elements in a custom element for the same effect. 

/cc @sorvell

prushforth:
I guess the people who are waiting on this don't mind working with polyfills. Is Mozilla planning on implementing html imports?  That will have to be polyfilled for the forseeable future, right?  Is there any plan to remove native builtins from the spec?  For my [purposes](http://maps4html.github.io/Web-Map-Custom-Element/blog/progressive-web-maps.html), wrapping a native element in a custom element doesn't give the same effect.  I gather this holds for others, but I'll let them speak for themselves!

seaneking:
AFAIK customized builtins are [in-progress](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/j7Oo9JSPY8c) in Chrome? There's also mentions in FF sourcecode about customized builtins, and Edge have said that they're 'basically okay' with moving on it too.

At the end of the day though `is=""` is in the spec, and whether it's put behind a flag in the polyfill or (better) only loaded when needed (ie: if type-extended elements are present), and the vendor support issue is documented, then I don't see the problem with polyfilling it? 

Then it's up to developers to determine whether they need to ship customized built-ins with the extra weight that may incur, rather than have that decision forced upon them upstream. A polyfill shouldn't be opinionated, IMO. Because there are still many use-cases that just aren't possible without extending built-ins, it's not an ideological argument.

And, IMO, if we stopped using something every time a browser vendor didn't ship it then we wouldn't have WC at all (eg: service workers and Safari).

/rant

seaneking:
And yes, following this reasoning HTML imports shouldn't polyfilled at all, as FF have refused to implement.

joeldenning:
As the author of this pr, I know that there is real risk that Safari will never implement customized built-in elements, but believe that it is worth having the polyfill meet the spec. See #42 for previous discussion, also https://twitter.com/Joelbdenning/status/839241923139268608 for some of the discussion and thoughts as I was deciding about making this pr or not.

In the end, I believe that as long as users of the polyfill are aware of the risks associated with using customized builtins, that they should be able to. See https://github.com/webcomponents/custom-elements/pull/88/files#diff-04c6e90faac2675aa89e2176d2eec7d8R84 for where we warn the user in the Readme. Also note that I have built this feature behind a feature flag, as discussed in #42.

treshugart:
> As the author of this pr, I know that there is real risk that Safari will never implement customized built-in elements, but believe that it is worth having the polyfill meet the spec.

+1

I wonder if it would be more palatable if it were offered as an opt-in fill. Meaning, the consumer has to deliberately include a separate polyfill for customised built-ins (that assumes standard custom element support is provided).

I think Safari currently has around 10% market share. Considering that customised built-ins are extremely useful, it might be sane to add support for it, especially since the spec has it.

joeldenning:
@treshugart Customized builtins already are opt-in in this pull request. You have to run `customElements.enableCustomizedBuiltins = true` in order to use them. The idea to have customized builtins be implemented as a whole separate file is one that was discussed briefly in #42 and #13, and it was decided that that was not necessary.

The only advantage I can think of to separating it into a different script is to try to reduce the bundle size for the polyfill for those who are not using customized builtins. Looking at the bundle size, though, the polyfill is 14.1kb right now and my pull request adds only 3.5kb. That difference is negligible, in my opinion, and not worth causing the extra nuisance of users having to include two separate files.

prushforth:
If its part of the spec it should be built into the bundle. I'm not sure what advantage having to activate it would have, possibly speed for non-builtin sites?  How much time would be saved I wonder. More switches and files causes more confusion DX.

prushforth:
Ping! What is the process for this to go from here to Polymer 2?  Looking forward to trying it out.

seaneking:
Any update on this? At least a confirmation on the consensus of if this will be merged into the polyfill or not, or if the community should start investigating forking/splitting it out. 

This is still a blocker for a lot of people in migrating to v1 specs / Polymer 2, would be great to have a final word either way from maintainers.

joeldenning:
## Status update

Last night we had a [small discussion on twitter](https://twitter.com/BedeOverend/status/872240905561841664) about what to do with this pr. It looks like I perhaps misunderstood the reasoning for splitting out the customized builtins into its own script. I thought it was to prevent people from using customized builtins accidentally without knowing the risks associated with this part of the spec. But it sounds like it also is motivated by a desire to use native custom element behavior for autonomous custom elements where possible (ie Chrome) and only use the polyfill code for customized builtins.

## Question for the reviewers:

* Am I understanding correctly what the motivation is for splitting out customized builtins into a separate script? To make it possible to only polyfill customized builtins but still use native autonomous custom elements?
* Is that the primary reason for creating a separate bundle for customized builtins? Or is the risk (that webkit never implements customized builtins) also a reason to do so?

## Implementation options:

Splitting things out into separate scripts creates some new implementation challenges. Specifically, in order for *any type of custom element to be polfyilled*, mutation observers must be set up and the window/DOM needs to be patched. So if we want two separate bundles (one for autonomous and one for customized builtins), a (maybe naive) implementation would create a situation where the mutation observers and window patching would happen twice if you used both scripts in your project. To me, that does not seem ideal.

With that said, here are the options I see for pulling off what we want:
1) Two separate bundles for autonomous and customized builtins, both setting up mutation observers and patching the window
2) One bundle for all of it (with only one set of mutation observers / window patches). But add some initialization logic that feature checks native implementations of custom elements to see if they support customized builtins. If the browser doesn't support it use the polyfill for customized builtins while still using native for autonomous.

I lean towards 2) because feature checking native behavior before polyfilling is something that is expected from a polyfill, plus doing it that way avoids the double mutation observer situation.

Which option do you prefer? Are there other options I am missing?

joeldenning:
cc @notwaldorf @justinfagnani. Would love to hear your response to the above comment!

seaneking:
Bump. 

Hate to be a nag, but would be great to know final stance from maintainers on how this PR is going to be treated (re: @joeldenning's comment above), so that the developers who rely on it to migrate to v1 specs can plan their work.

Is there anything we can do to help move this along? Do you need usage/use-case data of customized builtins? Perf testing? Anything?

justinfagnani:
@seaneking I haven't been the active maintainer here for a little bit, but in my mind we would ideally have three bundles/entrypoints:

1. Custom Elements w/o is=
2. Custom Element w/ is=
2. Only is=

This would allow libraries that depend on is= to document that they need the is= polyfill on Safari.

I would generally oppose adding is= directly to the main polyfill and only having one bundle, as I think it would encourage using is= without necessarily knowing that it required the full Custom Elements polyfill on Safari.

bicknellr:
@justinfagnani, that sounds like a good idea; splitting the polyfill's entry point into (1) and (2) would make the choice more explicit than a flag and would add less overhead for authors that aren't using `is=`. However, I'm not sure if (3) is feasible: is there a reasonable way to get correct timing for interleaved upgrades between elements using and not using `is=`?

@joeldenning, maybe we should look into how we could refactor `CustomElementInternals` (poorly named, sorry) and `CustomElementRegistry` objects so that the stuff that would need to be changed to support `is=` could be written as overrides in an extending classes? I.e. the new entrypoint for `is=` support would import these new extensions in place of the regular ones ([as `custom-elements.js` does here](https://github.com/webcomponents/custom-elements/blob/f27a9390f2827196f6649dc6ce19a9253d88bbef/src/custom-elements.js#L11)) and get the extra functionality without requiring explicit knowledge of this feature in the non-`is=` versions.

joeldenning:
@bicknellr @justinfagnani that sounds good. Creating three different bundles for those use cases will be a fair amount of work, but I think that it will help. I'll start work on it, and yeah maybe subclassing CustomElementInternals and/or CustomElementRegistry will be a good solution for this.

seaneking:
> I think it would encourage using is= without necessarily knowing that it required the full Custom Elements polyfill on Safari.

Isn't this what the `customElements.enableCustomizedBuiltins` flag (and its docs) is meant to communicate? My only concern is for users of webcomponents.js (which I assume wouldn't pull in `is=`), but if (3) is viable then its a moot point, extended components can just require that extra shim.

If `is=` requiring the whole polyfill is the primary blocker, would it make sense to ship this now and work on splitting out `is=` logic as a non-breaking change in future? Then affected devs can at least start migrating to v1 specs with current support.

bicknellr:
Yes, the flag certainly communicates that but I think there are a couple of things that make splitting into two bundles more desirable: users that don't intend to use `is=` won't pay any cost (or very little) for extra logic in terms of performance and size and, if `is=` ends up being dropped from the spec at some later point, removing it will be simpler. [1]

I think (3), the separate `is=`-only polyfill, would be ideal but I'm concerned that it's not actually possible to use the browser's custom element implementation to handle the non-`is=` elements while the polyfill manages the elements using `is=`, without a significant loss of ordering fidelity for upgrades and reactions. It might be worth exploring but I don't think we need to get stuck on it. That's why I think refactoring things a bit so that we can properly expose the points where `is=` needs special behavior seems like a more promising route.

[1]: I'm trying to gather whatever information I can but I'm in the dark and the implementor actions I'm aware of seem balanced. How do you compare a strong no against two partial but recently dormant implementations?

seaneking:
If (3) isn't feasible, and we still made separate bundles, then would webcomponents.js pull in the one that included `is=` behind a flag? If not then I think the minor savings in weight are far offset by the extra complexity.

snuggs:
> Anyway, IMO if we stopped using something every time a vendor didn't ship it then we wouldn't have WC at all - case in point: following this reasoning shouldn't HTML imports be dropped from polyfills (and Polymer) since FF have refused to implement? - @seaneking 

HTML imports IS (pun intended) "being dropped" somewhat conceptually ([if i'm reading this properly](https://github.com/w3c/webcomponents/issues/641)). Is a topic of discussion in the next TPAC meeting. As @notwaldorf states... perhaps unfortunately. /cc @bicknellr 
 https://github.com/w3c/webcomponents/issues/641

snuggs:
Also discussing here as well @seaneking @notwaldorf @bicknellr #StayWoke https://github.com/w3c/webcomponents/issues/645#issuecomment-311122827

prushforth:
Hopefully this is not forgotten in the Polymer Summit craziness.

tomalec:
@bicknellr  Do I understand correctly that you agree to merge implementation into the current code base/single bundle (behind the flag) once it will be somewhat separated internally and indicate clearly parts required only for `is`?

tomalec:
@joeldenning are you working on this? I would be happy to help, to push it sooner than later :)

justinfagnani:
"Getting this done" means locking some element users into the polyfill forever on Safari, and even Chrome for who knows how long. A polyfill that never goes away isn't really a polyfill. Getting off the pot, without creating a separate bundle with is= support, to me would simply mean closing this PR.

joeldenning:
I apologize for my absence. I hopefully will be able to work on this soon to create a separate bundle for is= support. @tomalec if you'd like to help out, let's collaborate! You can contact me on twitter at @joelbdenning.

fidele137:
Hi @joeldenning, is there any news about the customized built-in polyfill support?

idibidiart:
Thank you for having invested time in this. 

Chrome 67 ships with support for extending native elements. To leverage this feature in Chrome, we need to include this polyfill for other browsers. 

Any behind-the-scenes progress on this?

dfreedm:
This PR is very old, and now that we're moving to a monorepo, it has to be remade.

We have a tracking issue for this topic at https://github.com/webcomponents/polyfills/issues/102, would you mind remaking the PR in the monorepo?

Thanks!

internettrans:
@azakus agreed - this PR is very old.

I am no longer able to put time into this, but welcome someone else taking the lead on it.

PR #89, opened 2017-04-02
Updated the README.md to include install instructions

Pomax:
Added additional instructions for users interested in using this shim. These include:
- using npm
- cloning with git
- downloading a release
- manual download for the `.min` files.


Pomax:
Note that I cannot sign your CLA because it discriminates against people without a permanent address and phone, but as a README update rather than a code contribution, it should be pretty obvious there are no liabilities associated with this text update.

zigomir:
Nice, that is exactly what I was looking for. I'd maybe even add `unpkg` link one can include directly in html: `https://unpkg.com/@webcomponents/custom-elements` to get the latest version.

bicknellr:
Hi @Pomax, thanks for the improved instructions! Before we can bring this in, you'll need to [sign the CLA](https://cla.developers.google.com/).

bicknellr:
(test comment to see if i set up googlebot correctly)

Pomax:
hi @bicknellr
>Note that I cannot sign your CLA because it discriminates against people without a permanent address and phone, but as a README update rather than a code contribution, it should be pretty obvious there are no liabilities associated with this text update.

bicknellr:
Yeah, it doesn't seem like it would be a problem but I don't really know enough about our policy around that to make that call. :/ @tjsavage, do you know if I can pull this / what to do here?

Pomax:
worst case it's pretty generic text, filing the PR for a copy-paste from an already signed repo-contributor has the same effect (since the important part is a bit that explains how to do this, not the exact phrasing or who send the PR =)

Pomax:
I'm doing some repo cleanup so I'll be removing the fork I made specifically to file this PR, but I highly encourage you to either merge this, or write a quick "similar PR" to circumvent the need for CLA (I assume you've signed one ;) and merge that in.

Pomax:
small bump, as this should still be worked into the README. If not from my PR, then through an edit by someone who's already signed a CLA.

Pomax:
This is clearly not a PR that's ever going to land =)

PR #92, opened 2017-04-12
Make native-shim able to pass through a transpiler

doug:


samccone:
+1 would be super useful

justinfagnani:
Thanks @doug!

We've considered `new Function()` and `eval` approaches, but they're not CSP compliant. We'd have to offer the current version as well, complicating the decision tree of what polyfills to use.

Can I get an idea of why you want to pass the shim through a compiler? No non-ES6 supporting browser ever needs to load the shim, since they also don't support native Custom Elements. It it just so that the shim doesn't cause a syntax error in IE11, or is there some reason why it's difficult to exclude the shim from compilation?

doug:
That makes sense to be CSP compliant. I wanted to be able to run all of my js code through the closure compiler and get a single js binary. Rather than requiring loading of two js files in my HTML page. Not a big deal but just wanted to make my workflow cleaner. Wanted to be able to express fully all of my deps as es6 modules and pass that through closure or rollup or browserify. I can concat this shim onto my compiled js bundle or generate a file that evals this as a plugin to my build system, but it seemed less clean.

dgt41:
@doug you can transpile the code, here is what i've done:

```js
	var check = function () { try { eval("var foo = (x)=>x+1"); } catch (e) { return false; } return true; };

	// Check if ES6 then apply the shim
	if (check()) {
		// Trick so ES6 code won't throw in IE11
		(new Function("(()=>{'use strict';if(!window.customElements)return;const a=window.HTMLElement,b=window.customElements.define,c=window.customElements.get,d=new Map,e=new Map;let f=!1,g=!1;window.HTMLElement=function(){if(!f){const j=d.get(this.constructor),k=c.call(window.customElements,j);g=!0;const l=new k;return l}f=!1;},window.HTMLElement.prototype=a.prototype;Object.defineProperty(window,'customElements',{value:window.customElements,configurable:!0,writable:!0}),Object.defineProperty(window.customElements,'define',{value:(j,k)=>{const l=k.prototype,m=class extends a{constructor(){super(),Object.setPrototypeOf(this,l),g||(f=!0,k.call(this)),g=!1;}},n=m.prototype;m.observedAttributes=k.observedAttributes,n.connectedCallback=l.connectedCallback,n.disconnectedCallback=l.disconnectedCallback,n.attributeChangedCallback=l.attributeChangedCallback,n.adoptedCallback=l.adoptedCallback,d.set(k,j),e.set(j,k),b.call(window.customElements,j,m);},configurable:!0,writable:!0}),Object.defineProperty(window.customElements,'get',{value:(j)=>e.get(j),configurable:!0,writable:!0});})()"))();
	}
```

justinfagnani:
Unless someone like @azakus @kevinpschaaf @sorvell @notwaldorf objects, I think we should close this PR, since CSP compliance is a hard requirement for many of our customers.

doug:
Good with closing. CSP compliance is a perfectly reasonable constraint and easy enough to just concat this onto the front of a compiled script. Thanks.

PR #97, opened 2017-04-28
Workaround for Safari by using defineProperty to patch customElements

### LINKED ISSUE #96: Native shim fails to patch Safari intermittently
We've seen the native shim fail on Safari 10.1 where an element constructor that extended the patched `HTMLElement` is being run via a call to the native `customElements.define`, and not through the patched `customElements.define` (which ensures `constructorByTagname` is properly set).  This should be impossible given the code since both defining `window.HTMLElement` and patching `customElements.define` occur synchronously in the same script.  As such, it can probably be explained by a native object wrapper being garbage collected.

### Steps to reproduce:
1. npm install -g polymer-cli
2. mkdir test-app
3. cd test-app
4. polymer init (select polymer-2-starter-kit)
5. polymer build --compile-js
6. polymer serve build/default
7. reload in Safari until it throws (may take 20-50 reloads)

We believe we've found a workaround by using `Object.define` on `window.customElements` as well as `customElements.define` and `customElements.get` to do the patch.  With this change, we've let it run on an automated loop for >1000 reloads and haven't seen the issue).  Will submit PR for that.

kevinpschaaf:
Fixes #96.

### DIFF
--- a/src/native-shim.js
+++ b/src/native-shim.js
@@ -116,7 +116,7 @@
   // ptototype chain of built-in elements.
   window.HTMLElement.prototype = NativeHTMLElement.prototype;
 
-  window.customElements.define = (tagname, elementClass) => {
+  let define = (tagname, elementClass) => {
     const elementProto = elementClass.prototype;
     const StandInElement = class extends NativeHTMLElement {
       constructor() {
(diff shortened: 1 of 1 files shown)


kevinpschaaf:
Can't really add a good test since the failure is so intermittent...

src/native-shim.js
```diff
@@ -116,7 +116,7 @@
   // ptototype chain of built-in elements.
   window.HTMLElement.prototype = NativeHTMLElement.prototype;
 
-  window.customElements.define = (tagname, elementClass) => {
+  let define = (tagname, elementClass) => {
```
justinfagnani: const

src/native-shim.js
```diff
@@ -149,7 +149,16 @@
     constructorByTagname.set(tagname, elementClass);
     nativeDefine.call(window.customElements, tagname, StandInElement);
   };
-
-  window.customElements.get = (tagname) => constructorByTagname.get(tagname);
-
+  
+  let get = (tagname) => constructorByTagname.get(tagname);
```
justinfagnani: const

src/native-shim.js
```diff
@@ -149,7 +149,16 @@
...
+  
+  // Workaround for Safari bug where patching customElements can be lost, likely
+  // due to native wrapper garbage collection issue
+  Object.defineProperty(window, 'customElements',
```
justinfagnani: This is the strangest of the three lines - it should be a no-op. Is it required for the workaround?

kevinpschaaf: Yup, can't explain.

PR #102, opened 2017-06-15
Hold references to the pre-patched environment to prevent calling into later-applied polyfills.

bicknellr:
Most of this PR is leading whitespace changes:
https://github.com/webcomponents/custom-elements/pull/102/files?w=1


snuggs:
@bicknellr I notice pwe often don't check `window.`](https://github.com/webcomponents/webcomponentsjs/blob/master/webcomponents-loader.js#L22) in webcomponentsjs  for features which we should (Especially now that `const` and `let` create variables on the global space not the global Environment Record. (See [here](http://2ality.com/2013/09/window.html) and [here](ttp://tobyho.com/2013/03/13/window-prop-vs-global-var) Moving forward we actually should all be doing this check as

```javascript
const Element = 'foo'
window.Element !=== Element // true
```
Therefore we have to reassign `Element.prototype` After we upgrade to keep this polyfill from breaking.

I'll take care of that LOC in a PR. Let me know if you need help contributing to this PR/branch. 
capture d ecran 2017-06-21 a 18 38 08



- Peace!

bicknellr:
Hi @snuggs, I think it's very reasonable for the polyfill to assume that the built in environment works as it would by default. Without being able to make this assumption, the built in environment is rendered useless because any code intending to use it then needs to be aware of all code modifying it. I *strongly* recommend that you change the names of the objects you're adding to the global scope to prevent them from overwriting things that are built in.

The notable exception to this rule is a polyfill as they are specifically intended to modify the environment. However, a polyfill shouldn't be modifying the environment arbitrarily but rather only for the purpose of allowing other code to ignore incompatible implementations of the standard environment or changes to it over time (which, on the web, are almost always additions). Particularly, the reason why non-polyfill code shouldn't modify the environment is the same reason why a polyfill needs to modify the environment as minimally and compatibly as possible: any differences between the standard environment the polyfill intends to emulate and the environment as modified by the polyfill could potentially break other code using that piece of the environment.

snuggs:
> (Or convert to modules and avoid overwriting globals altogether! 

I don't think you get it. And i understand it's a new concept to you the Global Environment Record. But doesn't change my request to feature check on `window.` No globals are being overridden. Should run the following.

```
const HTMLElement = 'foo'
window.HTMLElement !=== Element // true
```

The irony is you are focused on ☝️ when [this very Polyfill does the same thing](https://github.com/webcomponents/custom-elements/blob/master/src/native-shim.js#L98).

You assumed the entire library isn't wrapped in a module at bundle time. I wanted to leave out packaging out of this conversation. Imagine it says `Element2`. Again. Checking on `window.Element` Is exactly what we should be doing. My library is superfluous to the conversation because we SHOULD be feature checking on `window`. If we are getting that into the weeds i'd even ask how the majority of Global feature checks are on `window.`.

You're setting up a strawman. Please focus [here](https://github.com/webcomponents/webcomponentsjs/blob/master/src/pre-polyfill.js#L63)
[here](https://github.com/webcomponents/webcomponentsjs/blob/master/src/post-polyfill.js#L15-L21)

bicknellr:
So, is it correct to say that you want this polyfill and others to reference all globals only as properties of `window`? If so, that sounds good to me - we should definitely remove any inconsistencies we find. Are there places where this polyfill doesn't do this that I've missed given the current state of this branch?

snuggs:
Oh no not at all @bicknellr we are DEFINITELY on the same page. Also should note I would never suggest something that I'm not willing to contribute with PRs myself :-) Just need a touch of guidance which actually this branch provided much of! I'm definitely not saying we should #ChangeAllTheThings however the platform is mature enough we can start closing the casket on some of the legacy things and should all be on the same page on how to do this.

Thanks for your feedback. This branch helped me on the "how" to contribute.

snuggs:
By the way @bicknellr one thing this conversation helped me clear up in my head that I would like your opinion on actually. How to bundle for "legacy" ES5 platforms. Although there is a new Global Environment Record. How can we still shield ourselves from leaving 💩 on a green pasteur.

One method i've been doing is utilizing variable shadowing via parameters. For instance we have a [`EventTarget` Mixin](https://github.com/devpunks/snuggsi/blob/master/elements/event-target.es) which takes care of the `this.on (event, handler)` syntactic sugar method _([think jquery](http://api.jquery.com/on/))_. The api is merely a wrapper around `EventTarget.addEventListener` with a render after event. the `Element` referenced within the mixin is actually a parameter passed in to the arrow function. Thus creating a closured shadow variable that doesn't mutate or taint `window.Element`. That said when a transpiler converts to ES5 that `EventTarget` mixin would then be declared with `var` thus clobbering `window.Element`. Now the concerns would rise that you speak of and you are 💯 correct!

The bundlers I use allow a module closure which allow wrapping the packet using the following convention:
```javascript
void function (GlobalEventHandler = window.GlobalEventHandler) => {

  /* polyfill */
  /* library */
  /* app */

} (/* a node, etc. `GlobalEventHandler` can be passed or default will be used */)
```
Just thought I would share some thoughts with you while we are on the topic.

bicknellr:
> That said when a transpiler converts to ES5 that EventTarget mixin would then be declared with var thus clobbering window.Element.

Yeah, that's the thing I was trying to get at earlier and I think your example at the end is the right way to go: you should be ok as long as the transpiled version of your library is run in its own closure. As far as 'exporting' in ES5 goes, assuming your users can't include specific fragments of your library through a module system, I think that adding a single global named after your library is as good as it gets - tons of libraries do this today. Without ES2015 modules or agreeing on a separate build-time module system, it's just the price of admission.

You can also set things up so there are two ways to include your library. For example, let's say you want to allow users to reference your code directly via `import` and without causing a global to show up but also want a pure-ES5 distribution that people can easily reference from a `<script>` at the top of their page, which has the global:

`Node.js`:
```js
export default /* definition */;
```

`Element.js`:
```js
import Node from './Node.js';
export default /* definition; uses your custom `Node` */;
```

`transpilation-root.js`:
```js
// Import all the things you want to expose.
import Element from './Element.js';
import Node from './Node.js';

// 'Export' them via one, uniquely named global. That's the best ES5
// really allows without building your own module thing on top. :/
window.snuggsi = {
  Element,
  Node,
};
```

Then, transpiling `transpilation-root.js` to ES5 for distribution will give you something with structure similar to this:

```js
(function() {
  var Node = /* definition of './Node.js' default export */;
  var Element = /* definition of './Element.js' default export, references `Node` above */;
  window.snuggsi = {
    Element: Element,
    Node: Node,
  };
}).call(self);
```

Rollup (which we use here to produce a debug build) produces output that gets structured almost exactly like that. Also, you can totally achieve this effect with a more concat-like build, too. I.e. if you're not using ES2015 modules in your source.

Your users writing ES5 can then reference all of the stuff you intended to export through `window.snuggsi` and do their own unwrapping if they would prefer shorter names:

```js
// A user of the ES5 distribution of your library probably assumes that your library
// has already run.
(function() {
  // This app might prefer that `Element` means your `Element` and not `window.Element`:
  var Element = window.snuggsi.Element;

  // .. actual application here ..
}).call(self);
```


Your users writing ES2015 can import things from `Node.js` and `Element.js` directly, if they want to reference both your `Element` and `window.Element`, they can decide if they want to import with a different name:

```js
// If they need both and want `Element` to be `window.Element`.
import snuggsiElement from '../path/to/your/Element.js';
```
```js
// If they only need yours or don't mind differentiating
// between `window.Element` and `Element`.
import Element from '../path/to/your/Element.js';
```

Also, by referencing your `Element`, etc. directly, `transpilation-root.js` won't be imported or bundled in their app, so your library's global (intended only for ES5 users) won't be set.

---

With regards to any mistakes you find in this branch: please feel free to point them out with comments on this PR or base your own PR off of it. :)

snuggs:
@bicknellr man you're good. And sometimes...life is as simple as `cat`. Took your "cat like" option a while ago. hehe. https://github.com/devpunks/snuggsi/blob/master/package.json#L40

bicknellr:
No magical insight here; I totally browsed through your project before writing that up. :)

snuggs (commented):
Nice commits @bicknellr well done! Quick question about https://github.com/webcomponents/custom-elements/pull/102/commits/dcf919d324f01a4288864096a1417bf99c875e8a. Any refs/repro/bug ticket/blog post relative to this? I'm learning about `HTMLTemplateElement` and it's implementation.  

FWIW a more terse operation would be:
```javascript
HTMLLinkElement.prototype,
window.HTMLTemplateElement && window.HTMLTemplateElement.prototype,
MutationObserver.prototype,
// etc
```

#### Example
capture d ecran 2017-09-22 a 10 26 06
 
Two means to the same end but I blame it on it being Friday ;-). If you do like the suggestion I can take care of the patch commit ASAP on this branch so you don't have to as a tip of the cap to awesome commits. 🎩👌 

And Happy Friday!

bicknellr:
Thanks! This branch has opened up a big can of worms both in terms of fixing the problem and testing the solution.

> Quick question about dcf919d. Any refs/repro/bug ticket/blog post relative to this? I'm learning about `HTMLTemplateElement` and it's implementation.

This commit is just in response to something I noticed while testing on IE11 that's only relevant to using the PatchMonitor thing in this branch, so there's not a bug about it out there. Is there specific info about `HTMLTemplateElement` that you're interested in?

> If you do like the suggestion I can take care of the patch commit ASAP on this branch so you don't have to as a tip of the cap to awesome commits.

Sounds good to me; if you open a PR with `env-capture` as the base branch with the change you mentioned (`window.HTMLTemplateElement ? window.HTMLTemplateElement.prototype : undefined` -> `window.HTMLTemplateElement && window.HTMLTemplateElement.prototype`), I'd be glad to merge it into this. You'll have to sign the CLA here first though: https://cla.developers.google.com/

Happy Friday to you too! :)

src/Environment/EventTarget.js
```diff
@@ -0,0 +1,17 @@
...
+export const prototype = constructor ? constructor['prototype'] : undefined;
+
+export const descriptors = !prototype ? {} : {
+  addEventListener: getDescriptor(prototype, 'addEventListener'),
+};
+
+const addEventListenerMethod =
+  method(descriptors.addEventListener) ||
+  // IE11
+  method(getDescriptor(window['Node'].prototype, 'addEventListener'));
```
snuggs: @bicknellr VERY slick! I <3 it.

bicknellr:
For reference, [this JSBin](https://jsbin.com/dexilel/1/edit?html,console) (forked from [this one](https://jsbin.com/niyeci/1/edit?html,console), produced by @robdodson, to force the polyfills) shows a symptom of the current polyfills' layering problems. In particular, the first element should be able to find the second during its `connectedCallback` because the element construction and insertion is atomic and all the `connectedCallback` reactions should happen afterwards. Currently, the polyfills call into each other incorrectly in such a way that each individual child node of the constructed result of the set to `innerHTML` is inserted in its own step, with any `connectedCallback` reactions being called before moving on to inserting the next child.

snuggs:
@bicknellr this explains so much! Been through this project a bunch of times and have to understand sometimes things are just the way they are just because!

bicknellr:
I didn't really include much description for the PR beyond the title; Rob recently ran across the issue being fixed here and showed me that link so I figured I'd add it here since for more context.

> things are just the way they are just because!

If you find something that isn't explained well, let me know and I'd be happy to answer questions and/or add docs as needed.

dfreedm:
Remaking this PR in the polyfills monorepo!

PR #117, opened 2017-09-30
Don't use arrow function, prevents IE11 syntax error

trusktr:


justinfagnani:
The shim is only intended to run on browsers which have native Custom Elements. On other browsers just use the polyfill. Loading both would impact performance more than necessary.

trusktr:
> The shim is only intended to run on browsers which have native Custom Elements. On other browsers just use the polyfill. Loading both would impact performance more than necessary.

Who are you speaking to? App authors, or library authors?

It should just be this: App author, stick library author's single script tag in your page and it will work regardless of browser.

web-padawan:
BTW Polymer 0.5 followed that single script approach and it used to work... sort of. But it was ES5. Things changed a lot since that time, now we need multiple builds: ES5 for IE11 and Safari 9 and ES2015 for Chrome and Safari.

trusktr:
@web-padawan Who makes those builds? Polymer? Or end user?

justinfagnani:
> Who are you speaking to? App authors, or library authors?

App authors, as they're the ones to know what browsers they support, whether they compile or not, and are responsible for installing any necessary polyfills.

> It should just be this: App author, stick library author's single script tag in your page and it will work regardless of browser.

It's never that simple with polyfills. The app author needs to install the polyfills. Libraries can't do it because they could be loading a polyfill when it isn't needed, or using a different polyfill than what the app author wants to use, or one that conflicts with a polyfill from another library.

trusktr:
A good polyfill is one that has a method for explicit definition (f.e. `polyfill.apply()`), and also one that gets out of the way when the existing polyfilled thing(s) exist(s).

Loading extra code is hardly the problem, people want an easy development experience across all browsers, especially single-person teams like hobbyists who don't want to worry about infrastructure as much, who just want to code an app and for it to just work.

A good loading experience up front can make the extra load time imperceptible.

trusktr:
Ah, f*** it. App authors, you're out of luck, use my lib only in Edge 12 and up.

That's one way to go! 😄

trusktr:
Move the web forward! 😆

trusktr:
@justinfagnani  I found a shortcoming of this adapter. I'm using it like this:

```js
// using my modified version, it uses eval() so that `class` syntax remains non-transpiled.
// it applies after detecting newer browsers only.
import "native-shim.js" 

// this class WILL be transpiled
class FooBar extends HTMLElement {}

customElements.define('foo-bar', FooBar)

let foo = new FooBar()

document.body.appendChild(foo)

foo.parentNode // TypeError: Illegal invocation.
```

This happens in an app of mine which is compiled with Buble. However, [running the code through Babel](http://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&code_lz=PQKhCgAIUghBDAzgUwCaQPYDtIAsAu-ADogFzDADmAlvrgK4BGAdAMYYC2wA7so-xyLZkWfImCt6ifJwC0yADbIOIscEYKMjYByT5kAJ2CIDrYFnj5qAN2SzEuah2YArRFBgAxDAcjUsAGYYkCjIeIQk5FS0DCwCPHwCQliq4rz8nMmpbgDEktJyisqp8ogArLLwqPBE-gZuHsDg4KAQ0JAAAgrUrCIoHpAAwhhEAJ4G1JQEkAAUrACUkABMAAwAjABskAAquGEAChgKoyq--wYYLsis-JAAgvR0PojM9woKkBNTYp_IKAa2VDMAa7aiISDsVBhXSjTBYY6QRhhKRoSD0LBQ3x0MKwADKABEQvhRkpIN1elgUJAguj0JZwsQokJjqdmDQ6ExmNQMMAADIASUGAFEAHK4oXMfAAD3wIL2EMySn0IWQtwwAUg8EeuGekBhiLCNIxmtuBEZFGZJ0MbJinO5wDuAFVtgAJADyACVcZKZXKwkklWEUGqNexRBNGI9dfqkdSMLSTQyiEyjlaDDaOSx7YM3SLth7-bBnZ7vdLZe1hlDIKgwfgI49UYxYQBxDAYSikpCQIjwAwhyDY7up07di5XG5-cHwBSIDADRBMce3GSanBVGtWbDTyD8_afSYEcGUAzwURxhP0s3Ji3D63s2Jcnn7O7bUXbUu-6BNcDUDUzbj-KgGDcGwUgyBwQpKCooiIIsADezSQEhCqUrcIqWDYyAutsACyvKQcUZ4ALyQABGLAcw2F4QR0HlshYbSJAFhWLY-LIAE_hhCRZFASB-TgTRqTMFCHEpFA9HYIxzGYc2qqQNxgEUfxnCCTBbKqohEmoQO8CUBYKiwKMwyoQY9A3D48lMcg3CQDhNQzPM4lIQxtwuaZ5kGIZ2y6fpXFWTZdlEA5zROZAFCQAA6ns2JYvKblmTIBhTgYYRIv4lAQtOSjoE2mqIhc3D_KOGC9IgiAADR-GEPZJeloXhRZrApZYyCqfgwLIWScmMAV_zGdI7mbjgJEBNO_T1cAkXRXssX-pJdYJbqvapcg6WZe8jawvAaL_PYPa9MVpUVVVE3rd0WAZaekBFLRKEDYt6ahUqO2GP1C03Nyw3UmNyCaUhPEUVR-FQaolkBOiH3YA5kAIZ1nW_rMACEPXAX182DZ98GhXDWmMfgPnwColn43phPIIZb2DT46n4DMdBgmw6MPY5OM4y530ANbIIMChIOCJHSbYsntawWX_opfFgSpIMwZVJO-Y52Os-FdlcyE9ApQOuD0oOo1q_FHnVhgfxYAA5K5WWIvArAcwOwT0-CBuJUrOMogYlMJZ9xOmb9rNw-z_jSKeB0C9Zsx69zvNlfMwV-51KX4BrOCB_jWC9C7kAAL4Z-FkFUkBTEYPTF2vKC4IAe8Bq_Nbew5bCQq4hU0gkv6UcoOCQF1X74VA21bBi4sqAa2t_i0NQ07UAAXhh2AZyjhWvUzkNfaNM6-8h2edeFhkqoQa2Dj2-CsLXkC9zL-Cm-CRAXDIxJENVFz332sIroOgvr0hPe4cDhEX1fN9F1GPfUi0J4BqwXJrOglhSCnU6kBSQtE2DNX0G1GYpt4Cm0WCnYOyB1Sn2_m1U63AfAc3BEiUWKI_CoRwXglE4Iz6_2YNfIugD76VW4I4Y-k44Raw_mFSatQWF339NrfwmANSRmoAofAsgxE3SEqFAGIEGGIOYbfIBfl0IsSwgQ8-TCAHCL-ndW4IlOJgwhkNOmBMVCVXkaIHmfMsZ-3ZnY_A5wWGWVcQ4sq-ihEaKMZ1Z6uJU6oH5FgNqlksD0HeAE5CyBrDThmAAAwzsE08oTwnn0sqwNu10ZQiFQOCLRmEVGg1hnHZyS9ErQ3KRUrek1BiWzfhhWw-DqJZKdj4OWjgjyYXBFIXhGdu6TVpIYWQTsvbYLTmELs5semmxgXUuGC4n7BSGcrSauwH5-OARXD4sZuAXAugaChVJBzzyKiidA8ZfAR3WTjcKOS-aVVnNSagUo_D4EWUs5CbpGBLmYMGdx6j75ugCHTHptjz7AowIrH5yEEYzERm7D2y8nHwrhirMBQYNZhCgbcN2shTEpHQJ03wjAi7gUvpgR4ds8pAVkFgIujgLr3Ljl_dpv9oaixiRiuGFzF4mU9tgb29A-F8vCo0yug5CXEtRGSnhNyqFB2md8vlSEvFt37u8CFYIWYYs3hilFVSvYjR-myw1ONLVwySXC1mUJIyUEoIYWJlTtIqoyTCyyaSMRhL7mo1h4rkI-oyX3LQ_xAR3EIPWfQ_Nrrn28S8cNhhI3RuoJGWNGcPVhJhYzLAKQbhoClQoRg1tbYkVcbmsMBb9CoGLaWm2WaQk5oAcJME1bri1vrWWzx0LW01kQB2wtdasoNo5k29JLaWHMEsHWdNDZBjawukW0dPaK19unbOmN3Ml3OpHe8MdE7fVYFzVUEYXbV0217b_U9QFagroPWW11OlSYGSMlU6mwYZiaueS-hWGcyVeWscgQFqorGvuQFC3-ib9U43fmxUSIGeUKHFuRSWBQILn2OvLMmlUQ1-vPrBrORj2bOtuCNCxn1wMKxhr8ROBgcCAdGN5CDNNqNk0WNnU6EUSG9njMaIIvhcTwFGhMRE9AMocMMNVSwx81rKUw7_R2V1YyaGkJVboXNjinSHni4I79SIniIE_SAlBeyludQqDay9JwLj4X8gFcr3FP2JKh3ilVTYKbaogU25UM5wQSQoMVpBSIS1Ahh7zlUwwcUoBreAGhkAhYWpBwztB4tKCSz7TORGHOduEuxTiznDCuaUeFgSWGPNyt8_5wLwXqwFZSFF7AMW4sJcy2K9hEx8ZtYHFlnL_y8tOcfsV0YbmlJS0U7RY6psyPVb9gF6cdWyNNcCJMVrGXesddS91jbyXsvgC408sqkBvAYAQL4ZA-SMT0N0b_GGh2Jvefy4htBQQMCyFLQYXzJ22zncVs9N7kSw6nfO2s-B9BEEUtQKMGdxmCmLqkagGYb3FYMSOCBzQlBkdtiYctUQIoMBQkWOFbYGihQGAuAYEL_INpmY-P4awJUZ5YGYEAA&debug=false&circleciRepo=&evaluate=false&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=6.26.0) works!

This is proving to be too difficult to maintain, I won't switch to Babel after having already switched to Buble.

---

Screw it. Sorry app devs, no IE 11 for you. Edge 12 and up. 😃

justinfagnani:
> Loading extra code is hardly the problem

We'll have to disagree to disagree here.

With the setup you just outlined you load the native-shim with your elements, slowing down the performance in Safari and Chrome, and increasing their payload unnecessarily.

As an app author I can decide whether or not I want to send ES5 and the native-shim to Chrome or just send ES2015. Personally I would choose the ES2015 since it's smaller and faster. If some elements hard-code an import of a polyfill or shim, I just wouldn't use them.

trusktr:
> As an app author

As a _highly experienced_ app author. You should consider people who want to learn web tech, and want to simply place a script tag that just works.

PR #127, opened 2017-12-15
Patch document.body setter

### LINKED ISSUE #126: Setting document.body does not call connectedCallback
### Background

I am using Rails with [Turbolinks](https://github.com/turbolinks/turbolinks). Custom Elements seem like a great fit for progressively enhancing server-side-rendered web apps (and are called out directly in the Turbolinks FAQ), so I tried using them with the help of this polyfill. Everything is great in Chrome with its native Custom Elements support, but in Firefox, I noticed that after moving to a new page the custom elements' `connectedCallback`s are not called, effectively making them useless to me.

### The Problem

When following a Turbolinks link, the library fetches the HTML response from the server and then simply replaces the document body, leaving the `head` mostly unchanged. It does this by setting `document.body`, pretty much like this:

```
function loadBody(htmlText) {
  const html = document.createElement('html');
  html.innerHTML = htmlText;
  document.body = html.querySelector('body');
}
```

`innerHTML` is patched by the polyfill, so custom elements are properly instantiated (the constructor is called). But there currently is no patch for the `document.body` property. This means that the polyfill doesn't noticed that the custom element was connected to the DOM, and thus `connectedCallback` is not called.

### Test Case

The following `wct` test shows the problem:

```html
<!doctype html>
<html>
<head>
<title>Document#importNode</title>
<script>
  (window.customElements = window.customElements || {}).forcePolyfill = true;
</script>
<script src="../../../../es6-promise/dist/es6-promise.auto.min.js"></script>
<script src="../../../../web-component-tester/browser.js"></script>
<script src="../../../custom-elements.min.js"></script>
<script>
function generateLocalName() {
  return 'test-element-' + Math.random().toString(32).substring(2);
}

function defineWithLocalName(localName) {
  const constructor = class extends HTMLElement {
    constructor() {
      super();
      this.constructed = true;
      this.connectedCallbackCount = 0;
      this.disconnectedCallbackCount = 0;
    }

    connectedCallback() {
      this.connectedCallbackCount++;
    }

    disconnectedCallback() {
      this.disconnectedCallbackCount++;
    }
  }

  customElements.define(localName, constructor);

  return constructor;
}

function createBody() {
  const body = document.createElement('body');
  const div = document.createElement('div');
  const element = document.createElement('custom-element');
  div.appendChild(element);
  body.appendChild(div);
  return body;
}

suite('Replacing document body', function() {
  let constructor;

  suiteSetup(function() {
    constructor = defineWithLocalName('custom-element');
  });

  setup(function() {
    const body = createBody();
    document.body.appendChild(body.firstChild);
  });

  test('Setting document.body to a new body with a custom element', function() {
    document.body = createBody();
    const element = document.body.querySelector('custom-element');
    assert(element instanceof constructor); // passes
    assert.equal(1, element.connectedCallbackCount); // fails
  });
});
</script>
</head>
<body>
</body>
</html>
```

denisw:
Fixes #126.

This PR ensures that `connectedCallback` / `disconnectedCallback` is called on custom elements if the body gets replaced by setting `document.body`. Notably, this improves compatibility with Rails' [Turbolinks](https://github.com/turbolinks/turbolinks), which re-sets `document.body` when following links.


denisw:
I signed it!

src/Patch/Document.js
```diff
@@ -71,6 +71,16 @@ export default function(internals) {
       return result;
     });
 
+  Object.defineProperty(document, 'body', {
+    get: function() {
+      return document.querySelector('body');
+    },
+    set: function(body) {
+      document.documentElement.replaceChild(body, document.body);
```
bicknellr: The reason this patch works is because [`Node.prototype.replaceChild` is patched](https://github.com/webcomponents/custom-elements/blob/6ad9993949a4f25c05b50b27a031a2ebfc49df89/src/Patch/Node.js#L122), which is being called here. I think it would be better if these patches weren't reentrant, `replaceChild` seems like a reasonable point to look for starters on how to implement this. I think this one will be a bit simpler because you don't have to worry about DocumentFragments (which will throw) or multiple nodes being inserted / removed at once (since it's a setter).

tests/html/Document/body.html
```diff
@@ -0,0 +1,89 @@
...
+<head>
+<title>Document#body</title>
+<script>
+  (window.customElements = window.customElements || {}).forcePolyfill = true;
+</script>
+<script src="../../../../es6-promise/dist/es6-promise.auto.min.js"></script>
+<script src="../../../../web-component-tester/browser.js"></script>
+<script src="../../../custom-elements.min.js"></script>
+<script>
+function generateLocalName() {
```
bicknellr: This doesn't seem like it's used anywhere.

src/Patch/Document.js
```diff
@@ -71,6 +71,16 @@ export default function(internals) {
       return result;
     });
 
+  Object.defineProperty(document, 'body', {
```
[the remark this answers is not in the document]
denisw: I didn't use `Document.prototype` because that didn't work. Now I found out why: `body` is actually defined on `HTMLDocument.prototype`. I'll update the PR accordingly.

src/Patch/Document.js
denisw: I thought this would be more elegant than duplicating code. But sure, I can also go the non-reentrant route here.

src/Patch/Document.js
```diff
@@ -71,6 +71,16 @@ export default function(internals) {
...
+    get: function() {
+      return document.querySelector('body');
```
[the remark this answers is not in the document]
denisw: Yeah, now that I know that the descriptor lives at `HTMLDocument.prototype` I can do this easily. 😅

tests/html/Document/body.html
denisw: True, I'm going to remove that.

denisw:
@bicknellr Thanks for the detailed review! I pushed an updated version that should address all of your comments. The patch is now applied to `Document.prototype` or `HTMLDocument.prototype`, depending on which is defined (the former on Chrome and the latter on Firefox, for instance).

src/Patch/Document.js
```diff
@@ -71,6 +71,28 @@ export default function(internals) {
...
+        }
+      }
+    });
+  }
+
+  if (Native.Document_body && Native.Document_body.get) {
+    patch_body(Document.prototype, Native.Document_body);
+  } else if (Native.HTMLDocument_body && Native.HTMLDocument_body.get) {
+    patch_body(HTMLDocument.prototype, Native.HTMLDocument_body);
+  }
```
bicknellr: Is this for IE / old Safari?

denisw: For Firefox. `Object.getOwnPropertyDescriptor(Document.prototype, 'body')` returns `null` on Firefox 58 Beta, but on `HTMLDocument.prototype` it works fine.

src/Patch/Document.js
```diff
@@ -71,6 +71,28 @@ export default function(internals) {
...
+      get: baseDescriptor.get,
+      set: /** @this {HTMLDocument} */  function(body) {
+        const replacedBody = this.body;
+
+        baseDescriptor.set.call(this, body);
+
+        if (body !== replacedBody) {
+          internals.disconnectTree(replacedBody);
+          internals.connectTree(body);
+        }
```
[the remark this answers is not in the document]
bicknellr: Actually, I'm not so sure about the ordering I mentioned here - still working on it.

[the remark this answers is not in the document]
bicknellr: Yup, the `replaceChild` wrapper is wrong: filed as #128. I think the spec text ordering would be preferable? (i.e. `internals.disconnectTree(body);` first)

src/Patch/Document.js
bicknellr: Huh, definitely not what I expected! I haven't found a bug report in their tracker - I'll file one if some more searching doesn't turn anything up.

src/Patch/Document.js
```diff
@@ -71,6 +71,28 @@ export default function(internals) {
       return result;
     });
 
+  function patch_body(destination, baseDescriptor) {
+    Object.defineProperty(destination, 'body', {
+      get: baseDescriptor.get,
+      set: /** @this {HTMLDocument} */  function(body) {
```
bicknellr: `HTMLDocument` &rarr; `Document`

src/Patch/Document.js
```diff
@@ -71,6 +71,28 @@ export default function(internals) {
...
+        const replacedBody = this.body;
```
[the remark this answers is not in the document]
denisw: Done.

src/Patch/Document.js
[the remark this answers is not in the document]
denisw: I  `internals.disconnectTree(body)` now if `body` is connected, as the first thing after the `baseDescriptor.set` call.

src/Patch/Document.js
denisw: I thought this made sense because there are other types of documents (e.g. `SVGDocument`) and `body` seems to be fairly HTML-specific. Or does the spec say something else?

bicknellr: Sorry about the delayed response. I've had a non-work obligation keeping me OOO the past couple of days which could potentially extend into next year (excluding Fridays). I'll also be taking next week (+ a couple more days) off. So, I probably won't be very responsive for most of the near future. :/

https://html.spec.whatwg.org/multipage/dom.html#the-document-object
AFAICT, it seems like HTML just extends `Document` from DOM and `HTMLDocument` is only meant to be a pointer to it for back-compat:
https://html.spec.whatwg.org/multipage/window-object.html#htmldocument
Huh... In total, the HTML spec contains "HTMLDocument" three times, once in the above link, and twice in reference to `createHTMLDocument()`. All six uses of "HTMLDocument" in the DOM spec refer to `createHTMLDocument()`. I always assumed that `HTMLDocument` played some more critical role in the HTML spec, maybe they found it unnecessary?

In Chrome and Firefox, `Object.getPrototypeOf(HTMLDocument.prototype) === Document.prototype` is true but also seems wrong since that second link reads to me as if it means `HTMLDocument === Document`. I haven't filed the Firefox bug yet but it looks like I might need to file one for Chrome too? This seems like either a case of the spec not reflecting a (possibly) unalterable implementations or a bug in, at least, both Chrome and Firefox.

bicknellr:
tests: go/webcomponents_custom-elements_pull_127
The only ones failing there are currently failing in master; looks good!

bicknellr:
Hmm, this is causing master (here) to fail on a lot of older browsers... investigating.

bicknellr:
Sorry, I had to revert this because it needs some work on Safari 9. The descriptor for `Document.prototype.body` isn't configurable there and I don't have a quick fix. Could you open a new request from the same branch?

denisw:
Oh, that's sad to hear. I created a new PR with the changes (#134).

Issue #115, opened 2017-09-22
`Node.prototype.textContent` does not have tests.


bicknellr:
https://github.com/search?q=textContent+path%3A%2Ftests%2Fhtml+repo%3Awebcomponents%2Fcustom-elements&type=Code

PR #130, opened 2018-01-09
fix(Node): setting textContent to null or undefined

limonte:
Fixes https://github.com/webcomponents/shadydom/issues/206

Also maybe fixes https://github.com/webcomponents/custom-elements/issues/115


TimvdLippe:
Hm, I might be missing the underlying reason. I would expect that calling `set` here with `undefined` would then call `set` as defined in `patch-accessors` in shadydom. Thus, the handling would be correctly set there. The tests in the shadydom PR you published did not fail, but for some reason two of the browsers are parsing `textContent` differently.

Is the underlying issue a race condition for which polyfill defines the accessor first (e.g. agnostic of the browser, but because of timing these two decided to fail)? In that case I do understand the reasoning, but would appreciate a clarifying comment added to this conditional to explain why.

limonte:
I can't say the exact reason why those two browsers are behaving differently. 

Two tests provided in this PR are failing in Chrome 41 for the same reason as in https://github.com/webcomponents/shadydom/issues/206 so maybe you can debug and understand the underlying issue.

PS. the build is failing not because of changes from this PR.

bicknellr:
I agree with @TimvdLippe's first sentiment. This doesn't seem related to the custom elements polyfill because the value given to the setter is passed directly to the one below it; it doesn't use the value in any other way.

limonte:
> This doesn't seem related to the custom elements polyfill because the value given to the setter is passed directly to the one below it

That makes sence. Sorry guys, but I don't have any better solution for this issue, so I'll close this PR.

> PS. the build is failing not because of changes from this PR.

Here's the fix for the broken build: https://github.com/webcomponents/custom-elements/pull/131

PR #134, opened 2018-01-12
Patch document.body setter

denisw:
Re-opening of #127 which had to be reverted.


bicknellr:
Thanks for opening this again. The issue that required reverting this earlier is that Safari 9 has a broken `Document.prototype.body` descriptor: it's both non-configurable and it's `get` and `set` are both undefined even though they exist (i.e. it isn't just a data descriptor). :/ So, fixing it will require (a) recreating the getter and setter (like `innerHTML` does) and (b) finding an appropriate place to put it.

As far as (b) goes, there are two options I can think of at the moment: either we introduce something like `CustomElementInternals`' `addPatch`[1] method for `Document`s or add the descriptor to `HTMLDocument.prototype` instead. `HTMLDocument.prototype` seems like the better of those two given that `body` is an extension to `Document` by the HTML spec and the distinction between `Document` and `HTMLDocument` seems to exist for back-compat even though they are (AFAICT) meant not to be separate objects. Also, [this section](https://github.com/webcomponents/custom-elements/pull/134/files#diff-bf436ab0481b11ac44130ab53d56583aR97) might end up being a bit more complex since we'll have to test one to decide if we should put it on the other.

What do you think?

[1] Right now, patches given to `addPatch` is only used in a couple of places and for a couple of scenarios, so it's not a huge problem, but I think [this patch](https://github.com/webcomponents/custom-elements/blob/76311bc2e7cd58b59ef641d6efa3536b05241f64/src/Patch/Element.js#L82) should really be guarding against `Node`s that aren't `Element`s so that it doesn't add descriptors to things that shouldn't have them. I should come up with a better way to describe this and file an issue.

bicknellr:
We merged #131 and #133 today which should get the tests to run properly again. You should be able to pick up the changes by merging master.

denisw:
Alright thanks, I rebased.

Thank you for the detailed explanation of what the problem is! I'll look into the possible approaches you listed and try them out to see what feels best.

dfreedm:
Please remake this PR in the new polyfills monorepo if this is still an issue for you: https://github.com/webcomponents/polyfills

Thanks!

PR #138, opened 2018-01-16
Fix #137: prefer `document.createElement` to `Document.prototype.createElement` for LastPass

### LINKED ISSUE #137: Custom Elements polyfill does not work with Last Pass password manager
This password manager appears to install an extension which overwrites `document.createElement`. The custom elements polyfill patches `Document.prototype.createElement` and therefore the Last Pass extension shadows this patch and custom elements created with `createElement` do not work.

One way to address this is to capture the method off the `document` instance rather than the prototype [here](https://github.com/webcomponents/custom-elements/blob/134c58e6b21a6788f18b715ac5d70ede0cad779f/src/Patch/Native.js#L2) and then delete `document.createElement` if it exists.

bicknellr:
I still need to actually try this out, please don't merge yet.

cc @usergenic; it won't let me add you as a reviewer for some reason.

### DIFF
--- a/src/Patch/Document.js
+++ b/src/Patch/Document.js
@@ -8,26 +8,36 @@ import PatchParentNode from './Interface/ParentNode.js';
  * @param {!CustomElementInternals} internals
  */
 export default function(internals) {
-  Utilities.setPropertyUnchecked(Document.prototype, 'createElement',
-    /**
-     * @this {Document}
-     * @param {string} localName
-     * @return {!Element}
-     */
-    function(localName) {
-      // Only create custom elements if this document is associated with the registry.
-      if (this.__CE_hasRegistry) {
-        const definition = internals.localNameToDefinition(localName);
-        if (definition) {
-          return new (definition.constructor)();
+  function patch_createElement(baseFunction) {
+    Utilities.setPropertyUnchecked(Document.prototype, 'createElement',
+      /**
+       * @this {Document}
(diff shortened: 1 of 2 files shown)


usergenic:
@bicknellr mabye because I'm not in webcomponents org?

bicknellr:
That's probably it; I think @azakus can add you if you want.

bicknellr:
I've tried this on my machine now: before this PR, installing the LastPass extension in Firefox causes most of the tests to fail; afterwards, it doesn't. Still needs to actually go through our CI though. @azakus mentioned that #131 would fix the tests; I'll ask him to confirm. Also, I think I need to pin the Promise polyfill to a particular version because they removed the `dist` directory (which contained the pre-built polyfill) from their repo recently.

bicknellr:
I need to change the replacement `HTMLElement` constructor to use the new `Document.prototype.createElement` also.

bicknellr:
> I need to change the replacement HTMLElement constructor to use the new Document.prototype.createElement also.
I fixed this by doing the repair before grabbing the descriptor that we consider to be built in.

TimvdLippe:
Is this PR ready for review to be merged? I know lots of teams willbe looking forward to this change.

lpellegr:
We fall on this PR while looking for solutions about the "common" issue with LastPass and Polymer. We look forward to this PR.

PR #143, opened 2018-02-05
Added support for native es6 classes in the native-shim.

### LINKED ISSUE #142: Native ES6 classes do not work when the native-shim is in use.
Currently if the native shim is in use, only es5 custom elements can be used.

**Console Output:**
`Uncaught TypeError: Class constructor MyElement cannot be invoked without 'new'`

RebeccaStevens:
Fixes #142

## Overview

Currently if the native shim is in use, only es5 custom elements can be used.
This pull request makes it so that es6 custom elements will work as well.

## Why is this need?

When working in a development environment, sometimes it is useful to keep the custom elements in their es6 forms. However, currently if the native-shim is being included, these elements will not work. One way to resolve this would be to dynamically import the native-shim based on whether you are in a development or production environment.

This pull request makes it so that is not need and everything "just works".


grimly:
Your changes would fail if a class uses a variable declared outside of the class definition but inside any non global scope, for instance an import would fail.

PR #146, opened 2018-02-19
Allow ES6 classes through native-shim.js

### LINKED ISSUE #142: Native ES6 classes do not work when the native-shim is in use.

grimly:
Fixes #142.

I only have Chrome and IE 11 available to run the tests, I completed them to make sure any issue is raised ASAP.

Reworked the native-shim.js to have a valid syntax for IE11 (one can only copy this shim and have no penalty on IE 11)

### DIFF
--- a/src/native-shim.js
+++ b/src/native-shim.js
@@ -55,31 +55,18 @@
 
   var NativeHTMLElement = window.HTMLElement, newHTMLElement;
 
-  try {
-    // Try first with new.target support. Use "eval" in case it does not exists.
-    newHTMLElement = eval(`(function(_super) {
-      return function HTMLElement() {
-        return _super(this, new.target || this.constructor);
-      }
-    })`)(_super);
-  } catch (e) { // Failed on new.target ?
-    newHTMLElement = function HTMLElement() {
-      return _super(this, this.constructor);
-    };
-  }
+  newHTMLElement = function HTMLElement() {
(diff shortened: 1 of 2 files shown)


RebeccaStevens:
Closes #143

bicknellr:
@justinfagnani is the most knowledgeable about the native shim, so I think he would really be the best reviewer here. If I remember correctly, the native shim specifically avoids `eval` (which both this PR and #143 use) so that it can be used with CSP types that block it.

justinfagnani:
Avoiding Reflect.construct is the main point of the shim being structured the way it is. At the time we measured, Reflect.construct was 20-30x slower than native super calls. If Reflect.construct has gotten faster enough since then to be viable, then I think that relying on the new Babel transform that allows extending built-in classes is a better options, since it eliminates the need for the shim altogether.

Also, we need to avoid eval(), which isn't CSP compliant.

grimly:
The shim actually breaks any web component that respect the actual specs, this is the main issue raised by #142.
If one component inserts it because he needs it, fine. It goes wrong when it breaks all other components. 

`Reflect.construct` is roughly 30% slower from my observations (Chrome 64).

I dropped the `eval()` and let `this.constructor` be the only option. This is not ideal but this is the best we have if we can't allow syntax errors.

PR #150, opened 2018-03-31
Implement `CustomElements.upgrade`

### LINKED ISSUE #148: Add support for `customElements.upgrade`.
https://html.spec.whatwg.org/multipage/custom-elements.html#dom-customelementregistry-upgrade

TimvdLippe:
I took the original tests from https://github.com/w3c/web-platform-tests/pull/9869 The last test upgrades inside a `template` content. However, at that point the elements did not have an `ownerDocument` yet. Therefore split up upgrading of an element into `upgradeElementWithDefinition` to force upgrading for these elements, when manually invoking `customElements.upgrade`.

Fixes #148


bicknellr:
Thanks for working on this!

I think that test from the web-platform-tests repo might be actually be incorrect in asserting that elements inside `template.content` would be upgraded after calling `customElements.upgrade(template.content)` though:

Here's the definition for `CustomElementRegistry#upgrade`:
https://html.spec.whatwg.org/multipage/custom-elements.html#dom-customelementregistry-upgrade
Which calls into 'try to upgrade an element':
https://html.spec.whatwg.org/multipage/custom-elements.html#concept-try-upgrade
Which calls into 'look up a custom element definition':
https://html.spec.whatwg.org/multipage/custom-elements.html#look-up-a-custom-element-definition
Which says "If document does not have a browsing context, return null.".

Also, `template` elements have an "appropriate template contents owner document", which is the owner document the `template.content` DocumentFragment and its descendants:
https://html.spec.whatwg.org/multipage/scripting.html#template-contents
The "appropriate template contents owner document" is created here:
https://html.spec.whatwg.org/multipage/scripting.html#appropriate-template-contents-owner-document
... in the steps that say "If doc does not yet have an associated inert template document, then:" "Let new doc be a new Document (that does not have a browsing context)."
So, the owner document for elements in `template.content` is a document without a browsing context. AFAICT, this means that "look up a custom element defintion" should return null and `CustomElementRegistry#upgrade` shouldn't upgrade them.

bicknellr:
Domenic just confirmed that the last bit of that test wasn't correct. I think this means `upgrade` only needs to be `this._internals.patchAndUpgradeTree(element);`.  :)

TimvdLippe:
@bicknellr Ah great, updated :+1:

tests/index.html
```diff
@@ -21,6 +21,8 @@
   WCT.loadSuites([
     'js/reactions.js',
     'js/registry.js',
+    // the next one is in a HTML file because it requires the template polyfill to be loaded
+    'js/registry-upgrade.html',
```
bicknellr: The top-level `js` and `html` test directories distinction is just left over from the layout before. I don't think those two were named in relation to what they were testing in any way; they seem to have originally been about grouping the tests by the format of the tests themselves. So, this test would make sense to go in `html`.

(FWIW, I'd be up for reorganizing these folders by what they're testing at some point, rather than what kind of tests they are.)

TimvdLippe: Feel free to push this branch to move the file. Else I can get to it tomorrow morning.

bicknellr: Ok, I'll make the update.

bicknellr:
Thanks again! I'll take a look at your other PRs soon.

bicknellr:
I think this qualifies as a minor version bump (post testing) since it's actually new functionality and not just an "implement this unsupported thing" bug.

PR #152, opened 2018-03-31
Add tests for `Node.textContent`

### LINKED ISSUE #115: `Node.prototype.textContent` does not have tests.

TimvdLippe:
Used the tests originally created by @limonte in #130 as well as a copy of the disconnect tests from `innerHTML`.

Fixes #115

### DIFF
--- a/tests/html/Element/index.html
+++ b/tests/html/Element/index.html
@@ -10,6 +10,5 @@
     './removeAttributeNS.html',
     './insertAdjacentElement.html',
     './insertAdjacentHTML.html',
-    './textContent.html',
   ]);
 </script>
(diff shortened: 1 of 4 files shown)


TimvdLippe:
Again Safari 9 breaking :cry:

bicknellr:
Looks like there is a bug Safari 9 where setting `textContent` to `""` actually causes the node's child nodes to be replaced with a single text node with text `""`, rather than just removing all the child nodes:
```javascript
div = document.createElement('div');
div.childNodes.length === 0; // true everywhere
div.textContent = "";
div.childNodes.length; // 1 in Safari 9, 0 everywhere else
```
I'd say change the values that you're setting to `textContent` to something other than the empty string; then, you can assert `div.childNodes.length === 1` after setting `textContent` in all cases.

TimvdLippe:
That fixed it, awesome thanks for the investigation @bicknellr !

TimvdLippe:
@bicknellr Done :+1:

PR #154, opened 2018-04-04
Add Reflect.construct based native shim

justinfagnani:
As an alternative to the current native-shim. This will be a bit slower, but easier to deploy as it's valid ES5 source and can be loaded on IE11 and modern browsers.

WIP, needs to be run through tests and we should check in a benchmark.


src/reflect-shim.js
```diff
@@ -0,0 +1,33 @@
...
+ * shim must only be loaded in browsers that do.
+ */
+(() => {
+  if (window.Reflect === undefined) {
+    // No Reflect, no classes, no need for shim
+    return;
+  }
+  const BuiltInHTMLElement = HTMLElement;
+  window.HTMLElement = function() {
+    return Reflect.construct(BuiltInHTMLElement, [], this.constructor);
```
bicknellr: `this.constructor` -> `new.target`

bicknellr: Oh, nevermind, this isn't set because we're assuming they've used `call` to get here... *facepalm*

justinfagnani: That, but mainly because `new.target` is a syntax error in older browsers.

TimvdLippe:
I suppose this supersedes #146? If so, you can use the tests from there (https://github.com/webcomponents/custom-elements/pull/146/files#diff-9235c84e2a0ddfda34371f7386d49564)

grimly:
BTW, while `super()` is 20 to 30 times faster than `Reflect.construct`, the loss in time spent is negligible (only +20% at most) on DOM element construction scale. I find no good reason to maintain two versions of this shim.

grimly:
Closes #146
Fixes #142

src/reflect-shim.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+  HTMLElement.prototype = Object.create(BuiltInHTMLElement.prototype, {
+    constructor: {value: HTMLElement, configurable: true, writable: true},
+  });
+  Object.setPrototypeOf(HTMLElement, BuiltInHTMLElement);
```
bicknellr: (from offline) `BuiltInHTMLElement` -> `Object.getPrototypeOf(BuiltInHTMLElement)`

grimly: This line is correct.
In order to perform a ES6 subclass A of B you need the following :

```javascript
Object.setPrototype(A.prototype, B.prototype);
Object.setPrototype(A, B);
```

The first is done differently with `Object.create`.

grimly: This line is correct.
In order to perform a ES6 subclass A of B you need the following :

```javascript
Object.setPrototype(A.prototype, B.prototype);
Object.setPrototype(A, B);
```

The first is done differently with `Object.create`.

bicknellr: Yup, just confirmed that you're right, thanks for the correction!

src/reflect-shim.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+ /**
+ * This shim allows elements written in, or compiled to, ES5 to work on native
+ * implementations of Custom Elements v1. It sets new.target to the value of
+ * this.constructor so that the native HTMLElement constructor can access the
+ * current under-construction element's definition.
+ *
+ * Because `new.target` is a syntax error in VMs that don't support it, this
+ * shim must only be loaded in browsers that do.
```
bicknellr: `new.target` isn't being used here; this comment doesn't seem to apply.

src/reflect-shim.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+(() => {
+  if (window.Reflect === undefined) {
+    // No Reflect, no classes, no need for shim
```
bicknellr: (i.e. You can't be using a native implementation of custom elements because it requires ES2015.)

src/reflect-shim.js
```diff
@@ -0,0 +1,33 @@
...
+ /**
+ * This shim allows elements written in, or compiled to, ES5 to work on native
+ * implementations of Custom Elements v1. It sets new.target to the value of
+ * this.constructor so that the native HTMLElement constructor can access the
+ * current under-construction element's definition.
+ *
+ * Because `new.target` is a syntax error in VMs that don't support it, this
+ * shim must only be loaded in browsers that do.
+ */
+(() => {
```
bicknellr: If we're assuming we can't use `new.target` / ES2015 syntax, then this arrow function probably shouldn't be here either, right?

bicknellr:
@grimly, I merged the test suite from your other PR into this one. Can you confirm here that what I did here is what you wanted? (Gotta make googlebot happy.) I went ahead and overwrote the old shim with this one. [Here's a version of a benchmark @justinfagnani wrote that I modified](https://jsperf.com/custom-elements-es5-shim-comparison) which includes the current shim; it looks like the `Reflect.construct` one is much better now. I also resolved my own comments; it would be helpful if you and @justinfagnani would review. :)

tests/html/shim.html
```diff
@@ -16,102 +16,299 @@
...
+  }
+  // Do nothing if the browser does not support Reflect API.
+  if (typeof Reflect !== 'object' || typeof Reflect.construct !== 'function') {
+    test.skip('This test suite does not apply to this browser (Reflect API missing).', () => {});
+    return;
+  }
+
+  test('works with createElement()', () => {
+    function ES5ReflectReflect1() {
+      return Reflect.construct(HTMLElement, [], this.constructor);
```
bicknellr: I think these need to be changed to `HTMLElement.call(this)`.

grimly: This was the intention.

Some transpilers (I don't remember which) use the Reflect API in the target code, then a polyfill make it an equivalent to `HTMLElement.call(this)` if the browser does not support it.
The goal of this test is to make sure the transpiled code still works on recent browsers

tests/html/shim.html
```diff
@@ -16,102 +16,299 @@
...
+    return;
+  }
+  // Do nothing if the browser does not support Reflect API.
+  if (typeof Reflect !== 'object' || typeof Reflect.construct !== 'function') {
+    test.skip('This test suite does not apply to this browser (Reflect API missing).', () => {});
+    return;
+  }
+
+  test('works with createElement()', () => {
+    function ES5ReflectReflect1() {
```
bicknellr: This looks like it was meant to be `ES5ReflectElement1`. Same goes for `...2` and `...3`.

tests/html/shim.html
bicknellr: Oh, I wasn't paying attention to the tests above which are already checking that `HTMLElement.apply(this)` works.

tests/html/shim.html
```diff
@@ -16,102 +16,299 @@
...
+suite('Native Shim with ES6 classes', () => {
+
+  // Do nothing if the browser does not natively support custom elements.
+  if (!window.customElements) {
+    test.skip('The native shim does not apply to this browser.', () => {});
+    return;
+  }
+
+  function createHTMLElementSubclassNamed(name) {
+    return eval('(function() { class ' + name + ' extends HTMLElement { constructor() { super(); }}; return ' + name + '; })')();
```
bicknellr: Given that this entire file is about testing the shim, which only matters if you have a native implementation of custom elements, maybe we should just go with ES6 here and switch at [the point that it's passed to WCT](https://github.com/webcomponents/custom-elements/blob/adcbc4f1ce75832191b03b211185029ea03b6fc0/tests/index.html#L33). Although, it is nice to see that tests are being skipped. Or, is this because polyserve (via WCT) is trying to compile this class out? I need to double check.

bicknellr: `polymer test` will compile out these classes if the browser it's testing on doesn't support them. However, if ES2015 isn't supported, then neither are native custom elements, so the checks in `suiteSetup` will skip the tests anyway. So, I've converted these to classes without using `eval`.

tests/html/shim.html
bicknellr: Fixed.

bicknellr:
I'm going to swap the CLA label since @grimly already confirmed this was the expected use of their code.

bicknellr:
Here's a benchmark comparing native custom elements using classes to ES5 with the original shim and ES5 using `Reflect.construct` in the shim: https://jsperf.com/custom-elements-es5-shim-comparison/1 . It includes adding a shadow root for the sake of doing something kind of realistic in the constructor, rather than just trying to stress the function call itself. For all the browsers it applies to, this seems like a pretty big win. (Chrome, Safari, Firefox with custom elements + shadow DOM manually enabled in about:config.)

bicknellr:
@justinfagnani ping

src/native-shim.js
```diff
@@ -10,155 +10,26 @@
...
-  // Workaround for Safari bug where patching customElements can be lost, likely
-  // due to native wrapper garbage collection issue
-  Object.defineProperty(window, 'customElements',
-    {value: window.customElements, configurable: true, writable: true});
-  Object.defineProperty(window.customElements, 'define',
-    {value: define, configurable: true, writable: true});
-  Object.defineProperty(window.customElements, 'get',
-    {value: get, configurable: true, writable: true});
-
+  Object.setPrototypeOf(HTMLElement.prototype, BuiltInHTMLElement.prototype);
```
justinfagnani: Previously this was:

```js
HTMLElement.prototype = BuiltInHTMLElement.prototype;
```

I think it should be still

justinfagnani:
Commits by not-me LGTM :)

src/native-shim.js
bicknellr: (from offline)
```js
HTMLElement.prototype = BuiltInHTMLElement.prototype;
HTMLElement.prototype.constructor = HTMLElement; // also needed
```

PR #156, opened 2018-04-19
Fix setting innerHTML on non-HTMLnamespaced elements

TimvdLippe:
Part of https://github.com/webcomponents/shadydom/issues/228 and required for https://github.com/webcomponents/shadydom/pull/229
The patch for `innerHTML` on Safari 9 called `createElement`, but for non-HTMLnamespaced elements this should be `createElementNS` with the `this.namespaceURI` instead.


src/Patch/Element.js
```diff
@@ -100,8 +100,14 @@ export default function(internals) {
...
           /** @type {!Node} */
-          const rawElement = Native.Document_createElement.call(document,
-            this.localName);
+          let rawElement;
+          if (this.namespaceURI === document.namespaceURI) {
+            rawElement = Native.Document_createElement.call(document,
+              this.localName);
+          } else {
+            rawElement = Native.Document_createElementNS.call(document,
+              this.namespaceURI, this.localName);
```
bicknellr: You can remove the if-else and use this line by itself: `createElementNS` can create elements in the HTML namespace too. (i.e. `document.createElement('button')` and `document.createElementNS('http://www.w3.org/1999/xhtml', 'button')` will both produce a regular HTML `<button>`.)

TimvdLippe:
@bicknellr Apparently Edge 14 `svg.innerHTML` is broken by itself, so we have no other option than to ignore the test :cry:

bicknellr:
What exactly is the behavior of Edge 14? Instead of disabling the test based on the UA string, could you check the behavior before the polyfill is loaded and decide from that? For example, in `insertAdjacentHTML`:

https://github.com/webcomponents/custom-elements/blob/fe10dc8a5b1936d839181fa65ded5903281d6c2b/tests/html/Element/insertAdjacentHTML.html#L6-L14

https://github.com/webcomponents/custom-elements/blob/fe10dc8a5b1936d839181fa65ded5903281d6c2b/tests/html/Element/insertAdjacentHTML.html#L171-L181

TimvdLippe:
@bicknellr Ah cool, yes that is better. Updated :+1:

tests/html/Element/innerHTML.html
```diff
@@ -165,14 +173,25 @@
...
+
+  const canSVGInnerHTML = (function() {
+    try {
+      // Edge 14 svg.innerHTML is broken even without the polyfill
+      const svg = NATIVE.createElementNS.call(document, SVG_NAMESPACE, 'svg');
+      NATIVE.innerHTML.call(svg, circleText);
+      return svg.querySelector('circle') instanceof SVGCircleElement;
+    } catch (e) {
+      return false;
+    }
```
bicknellr: Is this expected to throw in some cases? If not, you don't need to try-catch, IMO.

tests/html/Element/innerHTML.html
```diff
@@ -163,6 +172,26 @@
...
+  const canSVGInnerHTML = (function() {
+    // Edge 14 svg.innerHTML is broken even without the polyfill
+    const svg = NATIVE.createElementNS.call(document, SVG_NAMESPACE, 'svg');
+    NATIVE.innerHTML.call(svg, circleText);
+    return svg.querySelector('circle') instanceof SVGCircleElement;
+  })();
+
+  test('svg innerHTML', function() {
+    if (!canSVGInnerHTML) {
+      this.skip();
```
bicknellr: Oh, this is nice to know about. I think there are a few other places in the tests where I ended up switching out the test function itself:
```
const canDoTheThing = ... some test ... ;
(canDoTheThing ? test : test.skip)('test name', function() { ...
```
Calling `this.skip()` in them would be much cleaner.

PR #163, opened 2018-07-05
Polyfill Range methods which modify the dom.

### LINKED ISSUE #162: No callbacks when using range.insertNode
Using the polyfill on Edge, if I insert a custom node with range.insertNode, the connectedCallback never runs.
I tried creating the node with document.createElement or by new MyCustomElement, no callbacks run in either case.

ribrdb:
Fixes #162


bicknellr:
Wow, thanks for taking this on! This is a pretty big patch and I've got a handful of other things going on at the moment so it might take some time for me to review - apologies in advance.

bicknellr:
An update: I've been going through the specs to learn more about how ranges and custom elements work together and, AFAICT, all browsers with native implementations of custom elements (Chrome, Firefox Nightly, Safari) get the ordering for constructor and `connectedCallback`/`disconnectedCallback` calls wrong when doing `cloneContents` and `extractContents`.

So, I filed some bugs:
https://bugs.chromium.org/p/chromium/issues/detail?id=870464 (I don't have permission to swap to public but I've left a comment asking someone to do this.)
https://bugs.webkit.org/show_bug.cgi?id=188279
https://bugzilla.mozilla.org/show_bug.cgi?id=1480618

The polyfill should definitely use the spec-defined order, which roughly comes down to "do everything in one tree-order pass of the area contained by the range". I don't think this PR does that currently; I'll have to look through it some more. I guess the main point is that folks that using custom elements and ranges together will find that it's totally inconsistent from a timing perspective at the moment. :(

dfreedm:
There's actually a much larger problem with Ranges and the other polyfills as well: https://github.com/webcomponents/polyfills/issues/66

If you still need this, can you re-file this PR in the new monorepo: https://github.com/webcomponents/polyfills?

Thanks!

PR #164, opened 2018-07-13
Replace broken 'gulp debug' with 'npm run debug'

ruphin:
The gulp debug build command was broken and did not produce any output. I replaced it with a functionally identical npm run debug command that invokes Rollup directly to produce the same output.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,0 +1,6 @@
+const config = {
+  input: 'src/custom-elements.js',
+  output: { exports: 'named', file: 'custom-elements.min.js', format: 'iife', name: 'CustomElements', sourcemap: true }
+};
+
+export default config;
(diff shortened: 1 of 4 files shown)


rollup.config.js
```diff
@@ -0,0 +1,6 @@
+const config = {
```
bicknellr: I think you could just `export default {` instead of making the variable `config` - not important though.

rollup.config.js
```diff
@@ -0,0 +1,6 @@
...
+  input: 'src/custom-elements.js',
+  output: { exports: 'named', file: 'custom-elements.min.js', format: 'iife', name: 'CustomElements', sourcemap: true }
```
bicknellr: Could you remove the `exports: 'named',` and `name: 'CustomElements',` options here? We don't actually want things from the polyfill to show up anywhere except through the API it fakes (`window.customElements`, etc.). I tried running both with and without those options and it produced the same output on my machine, so I'm guessing they only apply when using some format that actually has named exports (i.e. not `iife`)? Either way, I think it's safer to remove these, unless there's another reason that I haven't thought of.

ruphin:
I added the requested changes

PR #169, opened 2018-08-05
Update Closure to latest version

TimvdLippe:
Also fix all warnings that were reported by Closure.
Most of the warnings were related to the types we defined
being incomplete.
This also picks up the `Document.readyState` extern definition of
https://github.com/google/closure-compiler/commit/c0a3465c9f5d84cb7e44d24419513b4e379d3e8a


externs/custom-elements.js
```diff
@@ -14,13 +14,13 @@ const CustomElementState = {
 /**
  * @typedef {{
  *  localName: string,
- *  constructor: !Function,
+ *  constructorFunction: !Function,
```
TimvdLippe: Closure did not like that this variable was called `constructor` and simply removed it from all type checks. Therefore, if you would create an object that would adhere to this interface, Closure would report the `constructor` should not be there, as the `constructor` of an object is:

```js
> ({}).constructor
< ƒ Object() { [native code] }
```

By renaming to `constructorFunction`, we can circumvent this.

src/Patch/Document.js
```diff
@@ -37,7 +37,7 @@ export default function(internals) {
      * @return {!Node}
      */
     function(node, deep) {
-      const clone = Native.Document_importNode.call(this, node, deep);
+      const clone = /** @type {!Node} */ (Native.Document_importNode.call(this, node, !!deep));
```
bicknellr: Could you comment that this `!!` (here and in `cloneNode` is because of a closure bug? (AFAICT, it thinks this isn't optional.) Also, I why it doesn't think `importNode` would return a `Node`? Is there a case where that can happen, excluding when it throws?

src/AlreadyConstructedMarker.js
```diff
@@ -4,6 +4,7 @@
  * element construction stacks.
  *
  * https://html.spec.whatwg.org/#concept-already-constructed-marker
+ * @extends AlreadyConstructedMarkerType
  */
 class AlreadyConstructedMarker {}
```
bicknellr: Not really important but I just realized we don't actually need a class + instance of it here:
```js
export default /** @type {!AlreadyConstructedMarkerType} */ ({});
```

src/Patch/Document.js
TimvdLippe: Also per MDN (https://developer.mozilla.org/en-US/docs/Web/API/Document/importNode):

> In the DOM4 specification, deep is listed as an optional argument. If omitted, the method acts as if the value of deep was true, defaulting to using deep cloning as the default behavior. To create a shallow clone, deep must be set to false.

> This behavior has been changed in the latest spec, and if omitted, the method will act as if the value of deep was false. Though It's still optional, you should always provide the deep argument both for backward and forward compatibility. With Gecko 28.0 (Firefox 28 / Thunderbird 28 / SeaMonkey 2.25 / Firefox OS 1.3), the console warned developers not to omit the argument. Starting with Gecko 29.0 (Firefox 29 / Thunderbird 29 / SeaMonkey 2.26)), a shallow clone is defaulted instead of a deep clone.

src/Patch/Element.js
```diff
@@ -268,7 +268,7 @@ export default function(internals) {
         } else if (position === "beforeend") {
           const marker = this.lastChild;
           baseMethod.call(this, position, text);
-          upgradeNodesInRange(marker || this.firstChild, null);
+          upgradeNodesInRange(marker || /** @type {!Node} */ (this.firstChild), null);
```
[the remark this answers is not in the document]
TimvdLippe: Done

TimvdLippe:
Ping @bicknellr I think this is ready for review and merge now 😄

PR #172, opened 2018-10-11
Improving documentation for how to use this project.

joeldenning:
I had a hard time using this project because the custom-elements.min.js file is what is described in the Usage docs, but it is not present in the Readme and I didn't know how to get that file without building it myself. It took some research and discovery to stumble across webcomponentjs and to realize that this github repo is published to npm under `@webcomponents/custom-elements` (and importantly *not* `custom-elements`).

So these changes will hopefully just help people not have a hard time using the project at first, like I did.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,12 +4,31 @@ A polyfill for the [custom elements](https://html.spec.whatwg.org/multipage/scri
 v1 spec.
 
 ## Using
+#### Standalone script tag
+Include a script tag at the beginning of your page, *before* any code that
+manipulates the DOM. The custom-elements.min.js can be loaded from a CDN:
+```html
+<script src="https://unpkg.com/@webcomponents/custom-elements"></script>
+```
 
-Include `custom-elements.min.js` at the beginning of your page, *before* any code that
-manipulates the DOM:
+Or you can build it yourself and then load it:
 ```html
 <script src="custom-elements.min.js"></script>
 ```
+
+#### Via npm:
+If you're using npm, webpack, or similar tools, this package can be installed and imported:
+```bash
+npm install @webcomponents/custom-elements
+```
+
+```js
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,13 +4,32 @@ A polyfill for the [custom elements](https://html.spec.whatwg.org/multipage/scri
...
 
+#### Via npm:
+If you're using npm, webpack, or similar tools, this package can be installed and imported:
+```bash
+npm install @webcomponents/custom-elements
+```
+
+```js
+// Do this import before any code that manipulates the DOM.
+import '@webcomponents/custom-elements'
```
TimvdLippe: I am not 100% sure this actually works. Have you verified it did? See a previous issue on this in https://github.com/webcomponents/webcomponentsjs/issues/994

tdumitrescu: Not the PR author, but I've used this pattern numerous times and it seems to work fine. Part of a standard polyfill module we use at Mixpanel looks like:
```js
import 'babel-polyfill';
import '@webcomponents/shadydom';
import '@webcomponents/custom-elements';
```
Slightly different (but open-source), in [the following example](https://github.com/mixpanel/panel-farm/blob/ae953f3bf82db4b14fff86fb7bb513d0425ab0d1/webpack.config.js#L43-L47) we let Webpack bundle the polyfills as top-level entrypoints to a single bundle called `polyfills`:
```js
const webpackConfig = {
  entry: {
    polyfills: [
      `@webcomponents/custom-elements`,
      `@webcomponents/shadydom`,
    ],
    main: `./src/index.js`,
    // etc
};
```
Both of these have been tested as working in browsers without any Custom Elements or Shadow DOM support.

TimvdLippe: Awesome, thanks for the confirmation. Merging :smile:

PR #177, opened 2018-11-05
Make custom elements compile with strict Closure conformance checks.

aomarks:


bicknellr:
Is there a way to enforce this here: https://github.com/webcomponents/custom-elements/blob/0d92cc4dba08c66d62f4b0c388a384327867c768/gulpfile.js#L22-L35 ?

aomarks:
> Is there a way to enforce this here: https://github.com/webcomponents/custom-elements/blob/0d92cc4dba08c66d62f4b0c388a384327867c768/gulpfile.js#L22-L35 ?

Probably, will need to investigate. I can take an action item to do that. For now we have them automatically tested in google3.

bicknellr:
SGTM. I think this repo might need to be updated to WCT 1.8.1 too: https://github.com/Polymer/tools/issues/751#issuecomment-433167805

PR #179, opened 2018-11-29
use typeof over instanceof in prototype and constructor checks

thecodejack:
Fixes issue#64


thecodejack:
Will create one more PR where CLA is done. Closing this.

PR #180, opened 2018-11-29
Replace with typeof over instanceof

### LINKED ISSUE #64: Compatibility with jsdom
Hello, I am having some troubles with jsdom. I found this bug tmpvar/jsdom#1737 that is related to this piece of code: https://github.com/webcomponents/custom-elements/blob/master/src/CustomElementRegistry.js#L64

It seems that lot of _sandboxing_ by jsdom is necessary to emulate the instanceof behavior.

This check is related to the specs? Every custom element class should really be instanceof Function?

thecodejack:
Fixes #64

### DIFF
--- a/src/CustomElementRegistry.js
+++ b/src/CustomElementRegistry.js
@@ -71,7 +71,7 @@ export default class CustomElementRegistry {
    * @param {!Function} constructor
    */
   define(localName, constructor) {
-    if (!(constructor instanceof Function)) {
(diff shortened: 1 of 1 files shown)


daniel-nagy:
Any reason not to merge this?

dfreedm:
Issue moved to https://github.com/webcomponents/polyfills/issues/105. If you still feel like this PR is important, please remake it in the polyfills monorepo.

PR #187, opened 2019-02-25
An exported variable can not be private

rictic:
Currently this has no meaning, but closure compiler will soon make this an error.


rictic:
Downstreamed as cl/235546994

PR #192, opened 2019-04-16
Separate patches for Node and Element; patch ShadowRoot with Node patches (i.e. `textContent`).

bicknellr:


bicknellr:
@sorvell; this has the comment fix now.

bicknellr:
Merging as @azakus says this passed internal testing.

PR #194, opened 2019-04-23
Prevent double-installation of es5 shim

rictic:
This may be more of an issue inside google due to the way that we've piped things through the build system, but this patch will ensure that even if the ES5 native shim is included in the page more than once, it is only installed once.

It's possible this is unnecessary in the wider world, in which case we can keep it as a downstream patch.

### DIFF
--- a/externs/custom-elements.js
+++ b/externs/custom-elements.js
@@ -62,3 +62,6 @@ Element.prototype.__CE_definition;
 
 /** @type {!DocumentFragment|undefined} */
 Element.prototype.__CE_shadowRoot;
+
+/** @type {boolean} */
+HTMLElement.es5Shimmed;
--- a/src/native-shim.js
+++ b/src/native-shim.js
@@ -16,13 +16,15 @@
  */
 (function() {
   if (
+    // An es5 shim is already loaded, don't load again.
+    HTMLElement.es5Shimmed ||
     // No Reflect, no classes, no need for shim because native custom elements
     // require ES2015 classes or Reflect.
     window.Reflect === undefined ||
     window.customElements === undefined ||
     // The webcomponentsjs custom elements polyfill doesn't require
     // ES2015-compatible construction (`super()` or `Reflect.construct`).
-    window.customElements.polyfillWrapFlushCallback
+    window.customElements['polyfillWrapFlushCallback']
   ) {
     return;
   }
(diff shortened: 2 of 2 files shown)


src/native-shim.js
```diff
@@ -16,13 +16,15 @@
...
+    // An es5 shim is already loaded, don't load again.
+    HTMLElement.es5Shimmed ||
     // No Reflect, no classes, no need for shim because native custom elements
     // require ES2015 classes or Reflect.
     window.Reflect === undefined ||
     window.customElements === undefined ||
     // The webcomponentsjs custom elements polyfill doesn't require
     // ES2015-compatible construction (`super()` or `Reflect.construct`).
-    window.customElements.polyfillWrapFlushCallback
+    window.customElements['polyfillWrapFlushCallback']
```
bicknellr: This also seems related to closure?

externs/custom-elements.js
```diff
@@ -62,3 +62,6 @@ Element.prototype.__CE_definition;
 
 /** @type {!DocumentFragment|undefined} */
 Element.prototype.__CE_shadowRoot;
+
+/** @type {boolean} */
+HTMLElement.es5Shimmed;
```
bicknellr: Why do we need this extern, given that the native shim isn't meant to be compiled? Assuming we do need it, can you prefix it with `__CE_`?

bicknellr:
```suggestion
HTMLElement.__CE_es5Shimmed;
```

src/native-shim.js
```diff
@@ -16,13 +16,15 @@
  */
 (function() {
   if (
+    // An es5 shim is already loaded, don't load again.
+    HTMLElement.es5Shimmed ||
```
bicknellr:
```suggestion
    window.HTMLElement.__CE_es5Shimmed ||
```

src/native-shim.js
```diff
@@ -33,5 +35,6 @@
   };
   HTMLElement.prototype = BuiltInHTMLElement.prototype;
   HTMLElement.prototype.constructor = HTMLElement;
+  HTMLElement.es5Shimmed = true;
```
bicknellr:
```suggestion
  HTMLElement.__CE_es5Shimmed = true;
```

dfreedm:
Let's pull this into the monorepo.

PR #195, opened 2019-04-26
Compile the native shim; fix `HTMLElement.name`.

bicknellr:
Closure Compiler seems to be removing the temporary object added in #193 that we need to make sure that `HTMLElement` continues to have the right `.name`. This PR uses `Object.assign` to assign `HTMLElement` to the window so that Closure Compiler can't remove the object.

cc @rictic


dfreedm:
Let's redo this in the monorepo

PR #196, opened 2019-04-30
Optimizations for creating/connecting elements

sorvell:
* When computing if an element isConntected, try to use `document.contains` before walking the dom tree. This is significantly faster, especially on Edge/IE.

* When an element is upgraded, check `hasAttributes` before iterating observedAttributes.

* When walking elements, create the Set of traversed imports lazily.

### DIFF
--- a/src/Utilities.js
+++ b/src/Utilities.js
@@ -29,10 +29,9 @@ export function isValidCustomElementName(localName) {
   return !reserved && validForm;
 }
 
-// IE11 does not have document.contains so polyfill it.
-if (!document.contains) {
-  document.contains = (node) => document.documentElement.contains(node);
-}
+// Note, IE11 doesn't have `document.contains`.
+const nativeContains = document.contains ? document.contains.bind(document) :
+  document.documentElement.contains.bind(document.documentElement);
 
 /**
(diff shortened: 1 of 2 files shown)


src/Utilities.js
```diff
@@ -39,7 +44,11 @@ export function isConnected(node) {
   if (nativeValue !== undefined) {
     return nativeValue;
   }
-
+  // Optimization: It's significantly faster here to try to use `contains`,
+  // especially on Edge/IE/
+  if (document.contains(node)) {
```
bicknellr: Given the `document.contains` polyfill above, would `node.ownerDocument.contains(node)` work in all cases here? I'd rather not polyfill that function if it isn't necessary.

src/Utilities.js
```diff
@@ -73,9 +82,9 @@ function nextNode(root, start) {
 /**
  * @param {!Node} root
  * @param {!function(!Element)} callback
- * @param {!Set<Node>=} visitedImports
+ * @param {?Set<Node>} visitedImports
```
bicknellr: Could we use `!Set<!Node>=` instead?

PR #198, opened 2019-05-23
Fix the `ChildNode#after` patching condition.

bicknellr:
Currently, `ChildNode#after` is patched when a native version of `ChildNode#before` exists.


azakus:
LGTM

PR #202, opened 2019-05-31
Adds a number of small optimizations

sorvell:
Setting customElements.preferPerformance (1) will use querySelectorAll to find elements to upgrade, connect, and disconnect, (2) does not use the document construction observer to upgrade the document. This setting improves performance but reduces correctness in some cases: (1) removes support for upgrading custom elements inside HTMLImports, (2) does not upgrade undistributed custom elements in polyfilled shadow roots, and (3) elements do not upgrade with Mutation Observer timing, which is less spec correct.

PR #203, opened 2019-05-31
Adds `polyfillLazyDefine`

### LINKED ISSUE #200: Add a `lazyDefine` feature to explore a future spec proposal.
Add a feature that allows defining custom elements via a function that is only run when the first element is found to be customized. This potentially allows the code needed to upgrade the element to be fetched on demand.

Something like `customElements.polyfillDefineLazy(name, classGenerator)` where `classGenerator` is a function that returns a promise that resolves to a custom elements constructor function.

sorvell:
Note, based on https://github.com/webcomponents/custom-elements/pull/202 which should be addressed before this.

Fixes #200 by adding customElements.polyfillDefineLazy(name, constructorGenerator) and adds a preferPerformance option.

The customElements.polyfillDefineLazy is an approximation of a potential spec feature to allow definitions to be processed lazily when the first element to customize is found. The constructorGenerator is a function that's invoked only the first time and element to customize is found. It can return an element constructor function or a promise that resolves to an element constructor function.

### DIFF
--- a/externs/custom-elements.js
+++ b/externs/custom-elements.js
@@ -15,6 +15,9 @@ CustomElementRegistry.prototype.forcePolyfill;
 /** @type {boolean|undefined} */
 CustomElementRegistry.prototype.preferPerformance;
 
+/** @type {function(string, !Function)|undefined} */
+CustomElementRegistry.prototype.polyfillDefineLazy;
+
 class AlreadyConstructedMarkerType {}
 
 /**
@@ -30,6 +33,14 @@ class AlreadyConstructedMarkerType {}
  * }}
  */
 let CustomElementDefinition;
+
+/**
+ * @typedef {{
+ *  localName: string,
+ *  constructorGetter: !Function
+ * }}
+ */
+let CustomElementLazyDefinition;
 
 
 // These properties are defined in the closure externs so that they will not be
--- a/externs/custom-elements.js
+++ b/externs/custom-elements.js
@@ -12,6 +12,9 @@
 /** @type {boolean|undefined} */
 CustomElementRegistry.prototype.forcePolyfill;
 
+/** @type {boolean|undefined} */
+CustomElementRegistry.prototype.preferPerformance;
+
 class AlreadyConstructedMarkerType {}
 
 /**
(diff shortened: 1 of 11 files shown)


externs/custom-elements.js
```diff
@@ -12,6 +12,12 @@
 /** @type {boolean|undefined} */
 CustomElementRegistry.prototype.forcePolyfill;
 
+/** @type {boolean|undefined} */
+CustomElementRegistry.prototype.preferPerformance;
```
bicknellr: BTW, there's still a few `preferPerformance` things in here.

bicknellr: (from offline) This branch is based on the other.

src/CustomElementInternals.js
```diff
@@ -38,6 +155,159 @@ export default class CustomElementInternals {
...
+      }
+    }
+  }
+
+  /**
+   *
+   * @param {string} localName
+   */
+  flushLazyDefinition(localName) {
+    const pendingLazyDefinition = this.localNameToLazyDefinition(localName);
```
bicknellr: `this._localNameToLazyDefinition.get(localName)`

src/CustomElementInternals.js
```diff
@@ -38,6 +155,159 @@ export default class CustomElementInternals {
...
+
+    // mark this definition in a pending state.
+    this.setLazyDefinition(localName, PendingLazyDefinitionMarker);
+
+    let constructorOrPromise;
+    try {
+      constructorOrPromise = pendingLazyDefinition.constructorGetter();
+    } catch (e) {}
+    if (!constructorOrPromise) {
+      return;
```
bicknellr: Maybe we should type check and throw?

bicknellr: Actually, I'm on the fence about throwing here. `connectTree` / `disconnectTree` don't currently catch errors even though [they're really supposed to be reported](https://html.spec.whatwg.org/multipage/custom-elements.html#invoke-custom-element-reactions).

sorvell:
Closed in favor of a simpler version that just does the synchronous optimization: https://github.com/webcomponents/custom-elements/pull/206.

PR #206, opened 2019-06-04
Adds `customElements.polyfillDefineLazy`

sorvell:
This is an optional optimization which allows defining an element with a function that returns a class. This function is called the first time the polyfill tries to upgrade an element of the given name. It allows the polyfill to defer any work needed to generate the element class or its callbacks until this work is actually needed.

### DIFF
--- a/externs/custom-elements.js
+++ b/externs/custom-elements.js
@@ -23,7 +23,7 @@ class AlreadyConstructedMarkerType {}
 /**
  * @typedef {{
  *   localName: string,
- * definition: !CustomElementDefinition
+ *   definition: !CustomElementDefinition
  * }}
  */
 let CustomElementDefinitionProducer;
--- a/externs/custom-elements.js
+++ b/externs/custom-elements.js
@@ -22,7 +22,7 @@ class AlreadyConstructedMarkerType {}
 
 /**
  * @typedef {{
- * localName: string,
+ *   localName: string,
  * definition: !CustomElementDefinition
  * }}
  */
--- a/externs/custom-elements.js
+++ b/externs/custom-elements.js
@@ -36,7 +36,7 @@ let CustomElementCallbacks;
  *  isComplete: boolean,
  *  localName: string,
  *  constructorFunction: !Function,
- *  callbacks: (CustomElementCallbacks|undefined),
+ *  callbacks: (!CustomElementCallbacks|undefined),
  *  constructionStack: !Array<!HTMLElement|!AlreadyConstructedMarkerType>,
  * }}
  */
@@ -76,3 +76,7 @@ Element.prototype.__CE_definition;
 
 /** @type {!DocumentFragment|undefined} */
 Element.prototype.__CE_shadowRoot;
+
+// Note, the closure type is incorrect here.
+/** @type {!HTMLCollection} */
+DocumentFragment.prototype.children;
--- a/externs/custom-elements.js
+++ b/externs/custom-elements.js
@@ -22,13 +22,21 @@ class AlreadyConstructedMarkerType {}
 
 /**
  * @typedef {{
- *  localName: string,
- *  constructorFunction: !Function,
  *  connectedCallback: Function,
  *  disconnectedCallback: Function,
  *  adoptedCallback: Function,
  *  attributeChangedCallback: Function,
  *  observedAttributes: !Array<string>,
+ * }}
+ */
+let CustomElementCallbacks;
+
+/**
+ * @typedef {{
+ *  isComplete: boolean,
+ *  localName: string,
+ *  constructorFunction: !Function,
+ *  callbacks: (CustomElementCallbacks|undefined),
  *  constructionStack: !Array<!HTMLElement|!AlreadyConstructedMarkerType>,
  * }}
  */
--- a/externs/custom-elements.js
+++ b/externs/custom-elements.js
@@ -22,21 +22,21 @@ class AlreadyConstructedMarkerType {}
 
 /**
  * @typedef {{
+ * localName: string,
+ * definition: !CustomElementDefinition
+ * }}
+ */
+let CustomElementDefinitionProducer;
+
+/**
+ * @typedef {{
+ *  localName: string,
+ *  constructorFunction: !Function,
  *  connectedCallback: Function,
  *  disconnectedCallback: Function,
  *  adoptedCallback: Function,
  *  attributeChangedCallback: Function,
  *  observedAttributes: !Array<string>,
- * }}
- */
-let CustomElementCallbacks;
-
-/**
- * @typedef {{
- *  isComplete: boolean,
- *  localName: string,
- *  constructorFunction: !Function,
- *  callbacks: (!CustomElementCallbacks|undefined),
  *  constructionStack: !Array<!HTMLElement|!AlreadyConstructedMarkerType>,
  * }}
  */
--- a/src/CustomElementDefinitionProducer.js
+++ b/src/CustomElementDefinitionProducer.js
@@ -145,7 +145,7 @@ export default class CustomElementDefinitionProducer {
       attributeChangedCallback,
       observedAttributes,
       constructionStack: []
-    }
+    };
 
     this._internals.setDefinitionConstructor(this._definition);
 
--- a/src/CustomElementDefinitionProducer.js
+++ b/src/CustomElementDefinitionProducer.js
@@ -15,7 +15,6 @@ let elementDefinitionIsRunning = false;
 export default class CustomElementDefinitionProducer {
 
   /**
-   *
    * @param {string} localName
    * @param {!Function} constructorOrGetter
    * @param {boolean} isGetter
--- a/src/CustomElementDefinitionProducer.js
+++ b/src/CustomElementDefinitionProducer.js
@@ -8,7 +8,7 @@
  * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
  */
 import CustomElementInternals from './CustomElementInternals.js';
- import * as Utilities from './Utilities.js';
+import * as Utilities from './Utilities.js';
 
 let elementDefinitionIsRunning = false;
 
--- a/src/CustomElementDefinitionProducer.js
+++ b/src/CustomElementDefinitionProducer.js
@@ -58,7 +58,10 @@ export default class CustomElementDefinitionProducer {
       throw new SyntaxError(`The element name '${localName}' is not valid.`);
     }
 
-    if (this._internals.localNameToDefinitionProducer(localName)) {
+    const existingProducer = this._internals.localNameToDefinitionProducer(localName);
+    // Note, only throw if the existing producer has a valid definition.
+    // If it does not, then re-defining is ok.
+    if (existingProducer && existingProducer.definition) {
       throw new Error(`A custom element with name '${localName}' has already been defined.`);
     }
 
@@ -70,11 +73,22 @@ export default class CustomElementDefinitionProducer {
     if (!this._isGetter) {
       this._ensureDefinition();
     }
+
+    this._internals.setDefinitionProducer(this.localName, this);
   }
 
   /** @return {!CustomElementDefinition} */
   get definition() {
-    return this._ensureDefinition();
+    try {
+      return this._ensureDefinition();
+    } catch (e) {
+      // If this definition is invalid, release it so it can be defined again.
+      this._internals.removeDefinitionProducer(this.localName);
+      // Note, if an error is generated while reading the definition, it's when
+      // the polyfill is in the processing of upgrading an entire tree and we
+      // don't want this to fail.
+      console.error(e);
+    }
   }
 
   /** @return {!CustomElementDefinition} */
@@ -92,6 +106,9 @@ export default class CustomElementDefinitionProducer {
     try {
       constructorFunction = this._isGetter ? this._constructorOrGetter() :
         this._constructorOrGetter;
+      if (!(constructorFunction instanceof Function)) {
+        throw new TypeError('Custom element constructors must be functions.');
+      }
       /** @type {!Object} */
       const prototype = constructorFunction.prototype;
       if (!(prototype instanceof Object)) {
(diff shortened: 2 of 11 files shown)


bicknellr:
Could you merge master again? That should cut down some of the diff.

src/CustomElementInternals.js
```diff
@@ -35,20 +35,82 @@ export default class CustomElementInternals {
     /** @type {boolean} */
     this._hasPatches = false;
 
+    /** @type {boolean} */
+    this.addingDefinitionCallbacks = false;
```
bicknellr: Assuming it's serving the same purpose, can we change this back to `elementDefinitionIsRunning`? This name is from the spec: https://html.spec.whatwg.org/multipage/custom-elements.html#element-definition-is-running

src/CustomElementInternals.js
```diff
@@ -35,20 +35,82 @@ export default class CustomElementInternals {
...
+
     /** @const {boolean} */
     this.shadyDomFastWalk = options.shadyDomFastWalk;
 
     /** @const {boolean} */
     this.useDocumentConstructionObserver = !options.noDocumentConstructionObserver;
   }
 
+  /**
+   * @param {Object} definition
```
bicknellr: `@param {!CustomElementDefinition} definition`

src/CustomElementInternals.js
```diff
@@ -35,20 +35,82 @@ export default class CustomElementInternals {
...
+        disconnectedCallback,
+        adoptedCallback,
+        attributeChangedCallback,
+        observedAttributes
+      }
+      this.addingDefinitionCallbacks = false;
+    }
+  }
+
+  ensureDefinitionComplete(definition) {
```
bicknellr: All the places where this is called seem to check immediately afterwards if the definition exists, as does this function. What if we change this to `localNameToCompleteDefinition` that would either return undefined if no definition exists or take the existing definition, complete it if necessary, and return the complete one?

src/CustomElementInternals.js
```diff
@@ -35,20 +35,82 @@ export default class CustomElementInternals {
...
+    if (definition && !definition.isComplete) {
+      try {
+        definition.constructorFunction = definition.constructorFunction();
```
bicknellr: I think that we should move `constructorFunction` to `CustomElementCallbacks` so that it isn't overloaded and replace it with a `constructorGetter` property specifically for that purpose. We should also consider renaming `CustomElementCallbacks` and `CustomElementDefinition`, since `CustomElementCallbacks` wouldn't be just callbacks anymore and `CustomElementDefinition` doesn't contain a full definition.

src/CustomElementInternals.js
```diff
@@ -35,20 +35,82 @@ export default class CustomElementInternals {
...
+    } catch (e) {
+      return;
+    } finally {
+      definition.callbacks = {
+        connectedCallback,
+        disconnectedCallback,
+        adoptedCallback,
+        attributeChangedCallback,
+        observedAttributes
+      }
```
bicknellr: `;`, I need to add a formatter to this repo. :/

externs/custom-elements.js
```diff
@@ -22,13 +22,21 @@ class AlreadyConstructedMarkerType {}
...
+ * }}
+ */
+let CustomElementCallbacks;
+
+/**
+ * @typedef {{
+ *  isComplete: boolean,
+ *  localName: string,
+ *  constructorFunction: !Function,
+ *  callbacks: (CustomElementCallbacks|undefined),
```
bicknellr: `(!CustomElementCallbacks|undefined)`

bicknellr (commented):
I think `CustomElementDefinition` should probably become a legit class rather than just a struct, given that this PR puts it in a position where it can have invalid states that can't be avoided using the type system. For example, `isComplete` should only be true if `callbacks` is not undefined.

externs/custom-elements.js
sorvell: Done

src/CustomElementInternals.js
sorvell: Done

src/CustomElementInternals.js
sorvell: Done

src/CustomElementInternals.js
sorvell: Done

src/CustomElementInternals.js
sorvell: Done

src/CustomElementDefinitionProducer.js
```diff
@@ -0,0 +1,136 @@
...
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+import CustomElementInternals from './CustomElementInternals.js';
+ import * as Utilities from './Utilities.js';
+
+let elementDefinitionIsRunning = false;
+
+/**
+ * @unrestricted
```
bicknellr: I don't think this should have `@unrestricted`. (AFAICT, the `@unrestricted` that's already on CustomElementRegistry shouldn't be there either.)

src/CustomElementDefinitionProducer.js
```diff
@@ -0,0 +1,136 @@
...
+    this._isGetter = isGetter;
+
+    /** @const {string} */
+    this.localName = localName;
+
+    /**
+     * @private
+     * @type {!CustomElementDefinition|undefined}
+     */
+    this._definition;
```
bicknellr: `= undefined`

src/CustomElementInternals.js
```diff
@@ -35,6 +35,9 @@ export default class CustomElementInternals {
     /** @type {boolean} */
     this._hasPatches = false;
 
+    /** @type {boolean} */
+    this.elementDefinitionIsRunning = false;
```
bicknellr: This isn't used.

src/CustomElementRegistry.js
```diff
@@ -142,6 +83,14 @@ export default class CustomElementRegistry {
     }
   }
 
+  define(localName, constructor) {
```
bicknellr, line 86: This needs types for the parameters.

src/CustomElementRegistry.js
```diff
@@ -142,6 +83,14 @@ export default class CustomElementRegistry {
...
+    this._define(localName, constructor, false);
+  }
+
+  polyfillDefineLazy(localName, constructorGetter) {
```
bicknellr, line 94: This also needs types.

src/CustomElementDefinitionProducer.js
```diff
@@ -0,0 +1,136 @@
...
+    // Per spec, reify the definition immediately if possible.
+    if (!this._isGetter) {
+      this._ensureDefinition();
+    }
+  }
+
+  /** @return {!CustomElementDefinition} */
+  get definition() {
+    this._ensureDefinition();
+    return /** @type {!CustomElementDefinition} */ (this._definition);
```
bicknellr: This cast can be avoided by making `_ensureDefinition` return the existing definition, if any, or generate the new one and return it and then changing this to `return this._ensureDefinition();`.

src/CustomElementDefinitionProducer.js
```diff
@@ -0,0 +1,136 @@
...
+      return;
+    }
+    elementDefinitionIsRunning = true;
+    let connectedCallback;
+    let disconnectedCallback;
+    let adoptedCallback;
+    let attributeChangedCallback;
+    let observedAttributes;
+    let constructorFunction;
+    let error;
```
bicknellr: `error` is unused.

src/CustomElementDefinitionProducer.js
```diff
@@ -0,0 +1,136 @@
...
+          throw new Error(`The '${name}' callback must be a function.`);
+        }
+        return callbackValue;
+      }
+      connectedCallback = getCallback('connectedCallback');
+      disconnectedCallback = getCallback('disconnectedCallback');
+      adoptedCallback = getCallback('adoptedCallback');
+      attributeChangedCallback = getCallback('attributeChangedCallback');
+      observedAttributes = constructorFunction['observedAttributes'] || [];
+    } catch (e) {
```
bicknellr, line 132: `define` is specced to re-throw exceptions that are thrown from inside any user code it calls<sup>1</sup>, in which case no definition is added to the registry's map.

So, you can do something like this:

```javascript
document.body.appendChild(document.createElement('custom-element'));

try {
  customElements.define('custom-element', class extends HTMLElement {
    constructor() {
      super();
      console.log("First definition constructed.");
    }
    static get observedAttributes() {
      throw new Error('Some error...');
    }
    // `observedAttributes` is only read if `attributeChangedCallback` exists. ^2
    attributeChangedCallback() {}
  });
} catch (e) {
  console.log('oops: ', e);
}

customElements.define('custom-element', class extends HTMLElement {
  constructor() {
    super();
    console.log("Second definition constructed.");
  }
});
```

... and the `<custom-element>` added in the first line will be upgraded with the second definition.

I think this means that this function should still bail when this catch happens and we'll need to update `get definition` to return `{!CustomElementDefinition|undefined}` as well as things that depend on it. Also, `this._internals.localNameToDefinitionProducer(localName)` doesn't completely determine if a definition already exists for an element anymore: you could have an existing producer for a lazy definition that will fail but hasn't been run yet. It would need to both check if the producer exists as well as force the producer to get the definition:

```javascript
const existingProducer = this._internals.localNameToDefinitionProducer(localName);
if (existingProducer && producer.definition) {
  throw new Error(`A custom element with name '${localName}' has already been defined.`);
}
```

We should pull the producer out of the map in CustomElementInternals when it fails in this way.

---

<sup>1</sup> Step 14 of [`define`](https://html.spec.whatwg.org/multipage/custom-elements.html#dom-customelementregistry-define) is run while catching any exceptions, then throws any caught exceptions as the last sub-step. The polyfill is incorrect and it suppresses these errors. This could probably be fixed by changing this `return;` to `throw e;`:
https://github.com/webcomponents/custom-elements/blob/92f66f713dc29e0428855ed7eaf4b1e5e02a3e92/src/CustomElementRegistry.js#L118

<sup>2</sup> The polyfill is incorrect and always reads `observedAttributes`:
https://github.com/webcomponents/custom-elements/blob/92f66f713dc29e0428855ed7eaf4b1e5e02a3e92/src/CustomElementRegistry.js#L116

bicknellr (changes requested):

src/CustomElementDefinitionProducer.js
```diff
@@ -0,0 +1,134 @@
...
+ * @license
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+import CustomElementInternals from './CustomElementInternals.js';
+ import * as Utilities from './Utilities.js';
```
bicknellr:
```suggestion
import * as Utilities from './Utilities.js';
```

src/CustomElementDefinitionProducer.js
```diff
@@ -0,0 +1,134 @@
...
+
+let elementDefinitionIsRunning = false;
+
+export default class CustomElementDefinitionProducer {
+
+  /**
+   *
```
bicknellr: extra line

src/CustomElementDefinitionProducer.js
```diff
@@ -0,0 +1,134 @@
...
+    this._definition = {
+      localName: this.localName,
+      constructorFunction,
+      connectedCallback,
+      disconnectedCallback,
+      adoptedCallback,
+      attributeChangedCallback,
+      observedAttributes,
+      constructionStack: []
+    }
```
bicknellr:
```suggestion
    };
```

tests/js/polyfill-define-lazy.js
```diff
@@ -0,0 +1,141 @@
...
+ * The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+suite('polyfillLazyDefine', function() {
+  var work;
+  var assert = chai.assert;
+  var HTMLNS = 'http://www.w3.org/1999/xhtml';
```
bicknellr: `HTMLNS` is unused here.

src/CustomElementDefinitionProducer.js
```diff
@@ -0,0 +1,134 @@
...
+    /** @const {string} */
+    this.localName = localName;
+
+    /**
+     * @private
+     * @type {!CustomElementDefinition|undefined}
+     */
+    this._definition = undefined;
+
+    if (!(constructorOrGetter instanceof Function)) {
```
bicknellr: We should copy this check into `_ensureDefinition` to test `constructorFunction`. Also, in this one, it might also be worth throwing errors with different messages depending on `isGetter` since the current error message still reads way it did when `constructorOrGetter` here was only `constructorFunction`.

externs/custom-elements.js
```diff
@@ -20,6 +20,14 @@ CustomElementRegistry.prototype.shadyDomFastWalk;
 
 class AlreadyConstructedMarkerType {}
 
+/**
+ * @typedef {{
+ * localName: string,
```
bicknellr:
```suggestion
 *   localName: string,
```

externs/custom-elements.js
```diff
@@ -20,6 +20,14 @@ CustomElementRegistry.prototype.shadyDomFastWalk;
...
+ * definition: !CustomElementDefinition
```
bicknellr:
```suggestion
 *   definition: !CustomElementDefinition
```

src/CustomElementInternals.js
sorvell: Removed.

src/CustomElementRegistry.js
sorvell: Done.

src/CustomElementRegistry.js
sorvell: Done.

src/CustomElementDefinitionProducer.js
sorvell: Removed.

tests/js/polyfill-define-lazy.js
sorvell: Removed.

dfreedm:
Ported to https://github.com/webcomponents/polyfills/pull/151

PR #213, opened 2019-06-07
Throw error when `createElement` does not produce an HTMLElement

### LINKED ISSUE #212: Error not thrown when createElement does not produce an element
When define is called with a constructor that does not extend HTMLElement, calling `document.createElement` for that name should fail.

sorvell:
Fixes #212.


src/CustomElementInternals.js
```diff
@@ -401,7 +409,104 @@ export default class CustomElementInternals {
...
+          }
+
+          return result;
+        } catch (e) {
+          this._reportTheException(e);
+
+          const result = /** @type {!Element} */
+              (Native.Document_createElementNS.call(
+                doc,
+                namespace === null ? NS_HTML : namespace,
```
bicknellr:
```javascript
namespace = namespace === null ? NS_HTML : namespace;
```
... at the top of the function would avoid three null checks.

src/CustomElementInternals.js
```diff
@@ -401,7 +409,104 @@ export default class CustomElementInternals {
...
+    // registry.
+    if (doc.__CE_hasRegistry && (namespace === null || namespace === NS_HTML)) {
+      const definition = this._localNameToDefinition.get(localName);
+      if (definition) {
+        try {
+          const result = new (definition.constructorFunction)();
+
+          if (result.__CE_state === undefined ||
+              result.__CE_definition === undefined) {
+            throw new Error("Failed to construct custom element.");
```
bicknellr: These errors should say something useful. Also, single quotes.

src/CustomElementInternals.js
```diff
@@ -401,7 +409,104 @@ export default class CustomElementInternals {
...
+  /**
+   * @private
+   * @param {!Error} error
+   * @see https://html.spec.whatwg.org/multipage/webappapis.html#report-the-exception
+   */
+  _reportTheException(error) {
+    const event = new ErrorEvent("error", {
+      cancelable: true,
+      error,
+    });
```
bicknellr: The other options from [`ErrorEventInit`](https://html.spec.whatwg.org/multipage/webappapis.html#erroreventinit) need to be filled out if possible. Unfortunately, it looks like the properties on `Error` for the file name and position might be UA-dependent.

src/Patch/Document.js
```diff
@@ -25,18 +26,7 @@ export default function(internals) {
...
-        if (definition) {
-          return new (definition.constructorFunction)();
-        }
-      }
-
-      const result = /** @type {!Element} */
-        (Native.Document_createElement.call(this, localName));
-      internals.patchElement(result);
-      return result;
+      return internals.createAnElement(this, localName, null);
```
bicknellr, line 29: I like being able to reuse this but I wonder if there's a way to guarantee that the native `createElement` is called when the wrapper for `createElement` is called, rather than always using the native `createElementNS`.

tests/html/HTMLElement/constructor.html
```diff
@@ -72,23 +69,89 @@
       });
     });
 
-  test('Custom element constructors may not cause more instances of the same ' +
-    'custom element to be created before calling into `HTMLElement` (e.g. via ' +
-    '`super()`).', function() {
+  test('Non-conformant custom element constructors that attempt to construct ' +
+    'more instances of themselves using `new` before calling into ' +
+    '`HTMLElement` (e.g. via `super()`) cause an error to be thrown when the ' +
```
bicknellr: This test description is now wrong.

bicknellr (dismissed):

src/CustomElementInternals.js
bicknellr: Constructing `ErrorEvent` doesn't work in older browsers.

src/Patch/Document.js
bicknellr: It looks like the template polyfill only works with `createElement`, making this necessary.

bicknellr: Fixed: passing `null` as the namespace now calls `createElement` instead of `createElementNS`.

src/CustomElementInternals.js
bicknellr: I updated this to create events using `createEvent` and `initErrorEvent` when `initErrorEvent` exists.

bicknellr:
This PR will fix https://github.com/webcomponents/polyfills/issues/112.

bicknellr:
Moved to https://github.com/webcomponents/polyfills/pull/139.

